PR #173, opened 2022-10-25
Adding --location to curl commands for install script

mikkelhegn:
To help prevent this in the future https://github.com/fermyon/spin/issues/858

Adding '-L' so if anyone copies the command to automation, redirects in the future will resolve.

Signed-off-by: Mikkel Mørk Hegnhøj <mikkel@fermyon.com>


bacongobbler:
Two more flags you could consider adding is

`-s`: silent mode
`-S`: show error even when silent mode is activated

In other words: `curl -sSL https://...`

We used those three flags in combination fairly frequently since Deis:

vdice (approved):

PR #176, opened 2022-10-25
check broken links

rajatjindal:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,3 +33,13 @@ jobs:
       - name: Check Docs
         run: |
           bart check content/* && bart check content/**/*
+
+      - name: Install spin
+        uses: engineerd/configurator@v0.0.8
+        with:
+          name: "spin"
+          url: "https://github.com/fermyon/spin/releases/download/v0.6.0/spin-v0.6.0-linux-amd64.tar.gz"
+          pathInArchive: "spin"
+
+      - name: Check broken links
+        run: .build/check-broken-links.sh
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,13 @@ jobs:
         with:
           node-version: 16
 
+      - name: Install spin
+        uses: engineerd/configurator@v0.0.8
+        with:
+          name: "spin"
+          url: "https://github.com/fermyon/spin/releases/download/v0.6.0/spin-v0.6.0-linux-amd64.tar.gz"
+          pathInArchive: "spin"
+
       - name: Install npm packages
         run: |
           npm ci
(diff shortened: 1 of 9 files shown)


rajatjindal:
- [ ] Need to use released version of `bartholomew` wasm file

.github/workflows/build.yml
```diff
@@ -32,4 +32,33 @@ jobs:
...
+          bart check content/* && bart check content/**/*
+
+      - name: Install spin
+        uses: engineerd/configurator@v0.0.8
+        with:
+          name: "spin"
+          url: "https://github.com/fermyon/spin/releases/download/v0.6.0/spin-v0.6.0-linux-amd64.tar.gz"
+          pathInArchive: "spin"
+
+      - name: Check broken links
```
karthik2804, line 44: Could we add this test to `package.json` as well? Potentially run the checks using `npm test` or `npm run test`. I think it would be a good idea to be able to test locally before pushing it to GitHub.

karthik2804 (changes requested):
LGTM overall but just one suggestion about adding tests for local environment

.github/workflows/build.yml
rajatjindal: as the broken-link-checker command line is long (due to the exclusion url's we have), do you think it might be good time to add either a script that is triggered by npm test (.build/check-broken-link.sh)  OR may be a Makefile?

karthik2804: Since we already have scripts to generate the search index, we could go down the script route but I do not have a strong preference though.

rajatjindal: ok, i have added the link checker script to `.build` folder.

package.json
```diff
@@ -37,6 +39,7 @@
     "spin": "nodemon --watch content --watch static --watch templates --ext md,rhai,hbs,css,js --verbose --legacy-watch --signal SIGINT --exec 'npm run build-index && spin up --file spin.toml'",
     "styles": "parcel watch static/sass/styles.scss --dist-dir static/css",
     "test": "npx markdownlint-cli2 content/**/*.md \"#node_modules\"",
-    "build-index": "node md_parser.mjs --dir=./content/ --out=./static/data.json"
+    "build-index": "node md_parser.mjs --dir=./content/ --out=./static/data.json",
+    "check-broken-links": ".build/check-broken-links.sh"
```
karthik2804, line 43: Can we add this to the the `npm test` command? So that it runs automatically along with markdown linter. Thanks!

karthik2804 (approved):
LGTM! Thanks :)

package.json
rajatjindal: done

Issue #184, opened 2022-10-28
Search feature breaks if lunr is not able to load


mikkelhegn:
We had an issue where the search feature was not loading because a script we depend on is not loading (HTTP 520): https://unpkg.com/lunr/lunr.js.

Suggest we consider hosting dependencies ourselves.

karthik2804:
closed with #186

PR #186, opened 2022-10-31
[#184] load script libraries locally

flynnduism:
Small edit to address #184. Changes how two javascript libraries are loaded - fetching them locally instead of via CDN.


karthik2804 (approved):
LGTM!

PR #189, opened 2022-11-02
added styling to TOC and float on right for large screens

karthik2804:
image


flynnduism (approved):
Thanks @karthik2804 this responds well to the browser sizing. 

* Is it possible to add a title to give context to what the box is? _Contents_ for example.
* Ideally, the positioning should be in relation to the other menus. Something like this. 

Frame 3


I think what's here is good enough to merge, I can follow up with some of the grid alignment as there are some broader issues with the layout I can tweak.

karthik2804:
@flynnduism Thanks for the review. Added the tags to indicate what the box is for. I will defer the grid alignment to you.

mikkelhegn:
Is this intentional on not so wide windows?
image

karthik2804:
> Is this intentional on not so wide windows?

Yes, it is intentional for not-so-wide windows.

PR #192, opened 2022-11-02
Add update search index step to PR template

karthik2804:


mikkelhegn (commented):
Should we add to the list that the data.json file should be committed after having run this step? Or should we just do this in the deployment pipeline?

vdice:
> Or should we just do this in the deployment pipeline?

Yes, I think we should at least try removing as many things as possible from being a burden to the dev/contributor to being handled by CI, this being one of them.  Also, the checklist mentions running other items that are either in CI automation on PRs or could be, like `bart check` and `npm run test` -- I'd vote if they are covered in PR CI, we remove them from the manual checklist.

karthik2804:
closed with #191  as it was added to the template.

PR #210, opened 2022-11-18
Corrected the route to match example code

robotastic:
The Rust Example code uses the /wildcard route instead of /hello. Updating the docs to match the code.

https://github.com/fermyon/spin/blob/013059cb7eda1189190fdec9fced86c9fc88867f/examples/http-rust-outbound-http/spin.toml#L22

Signed-off-by: Luke Berndt <lukekb@gmail.com>

- [ ] Has passed [`bart check`](https://bartholomew.fermyon.dev/quickstart)

### DIFF
--- a/content/spin/rust-components.md
+++ b/content/spin/rust-components.md
@@ -106,12 +106,12 @@ id = "hello"
 source = "target/wasm32-wasi/release/spinhelloworld.wasm"
 allowed_http_hosts = [ "some-random-api.ml" ]
 [component.trigger]
-route = "/hello"
+route = "/wilcard"
 ```
 
 Running the application using `spin up --file spin.toml` will start the HTTP
 listener locally (by default on `localhost:3000`), and our component can
-now receive requests in route `/hello`:
+now receive requests in route `/wildcard`:
 
 <!-- @selectiveCpy -->
 
--- a/content/spin/rust-components.md
+++ b/content/spin/rust-components.md
@@ -111,7 +111,7 @@ route = "/outbound"
 
 Running the application using `spin up --file spin.toml` will start the HTTP
 listener locally (by default on `localhost:3000`), and our component can
-now receive requests in route `/wildcard`:
+now receive requests in route `/outbound`:
 
 <!-- @selectiveCpy -->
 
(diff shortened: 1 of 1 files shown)


content/spin/rust-components.md
```diff
@@ -106,12 +106,12 @@ id = "hello"
 source = "target/wasm32-wasi/release/spinhelloworld.wasm"
 allowed_http_hosts = [ "some-random-api.ml" ]
 [component.trigger]
-route = "/hello"
+route = "/wilcard"
```
itowlson, line 109: I think the code here more closely follows the `/outbound` example (https://github.com/fermyon/spin/blob/013059cb7eda1189190fdec9fced86c9fc88867f/examples/http-rust-outbound-http/spin.toml#L11) - suggest using that rather than `wildcard`

content/spin/rust-components.md
```diff
@@ -106,12 +106,12 @@ id = "hello"
...
 ```
 
 Running the application using `spin up --file spin.toml` will start the HTTP
 listener locally (by default on `localhost:3000`), and our component can
-now receive requests in route `/hello`:
+now receive requests in route `/wildcard`:
```
itowlson, line 114: The bash sample on line 119 also needs to be reflect whatever change we decide to make.

content/spin/rust-components.md
tpmccallum, line 109: Thanks @robotastic for aligning these docs more closely with the actual Rust code that is offered. As @itowlson suggests, I also agree that we could/should call this outbound (instead of wildcard). Thanks for updating this from the generic `hello`. You are correct in saying that there needs to be a change/update. Thanks for your input and effort.
```suggestion
route = "/outbound"
```

content/spin/rust-components.md
tpmccallum, line 114:
```suggestion
now receive requests in route `/outbound`:
```

tpmccallum (changes requested):

tpmccallum:
Thanks for your input on this @itowlson 
The `outbound` text was a great idea. Also, the bash code on line 119 has been updated to suite.
https://github.com/fermyon/developer/pull/210/commits/170395c004dcecec7bf98a7258e26fc5bda48a2a

content/spin/rust-components.md
tpmccallum: Agreed, thanks Ivan.

content/spin/rust-components.md
tpmccallum: Done, in the second commit. Nice catch.

tpmccallum (approved):

itowlson (approved):

PR #237, opened 2022-12-16
update install.sh to support Linux ARM64

mikkelhegn:
Signed-off-by: Mikkel Mørk Hegnhøj <mikkel@fermyon.com>

- Removed the case where we say we don't support aarch64 on linux-gnu
- curl error is not only 'file not found' but any curl error, e.g. cannot write file output.

### DIFF
--- a/downloads/install.sh
+++ b/downloads/install.sh
@@ -89,10 +89,6 @@ esac
 case $OSTYPE in
 "linux-gnu"*)
     OS="linux"
-    if [[ $ARC == "aarch64" ]]; then
-        fancy_print 1 "The Processor type: ${ARC}, on ${OSTYPE} is not yet supported by Spin."
-        exit 1
-    fi
     ;;
 "darwin"*)
     OS="macos"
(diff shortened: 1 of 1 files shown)


radu-matei (approved):
LGTM, although I don't have a Linux ARM machine to test this.

cc @jpflueger

downloads/install.sh
```diff
@@ -89,10 +89,6 @@ esac
 case $OSTYPE in
 "linux-gnu"*)
     OS="linux"
-    if [[ $ARC == "aarch64" ]]; then
```
radu-matei, line 92: nit: `$ARCH` I think is the common variable name for the architecture.

mikkelhegn: Why did you comment on the code being deleted?

mikkelhegn:
> LGTM, although I don't have a Linux ARM machine to test this.
> 
> cc @jpflueger

Lima-VM on Macs with M1 or 2 is what I do.

PR #264, opened 2023-01-31
chore(install.sh): bump spin version strings to more recent value

vdice:
Although only informational, thought we'd bump the sample spin version strings to a more recent value.

[v0.8.0 is due to be released soon](https://github.com/fermyon/spin/pull/1092); we can wait for the release to finish before merging this


vdice:
Fix for build _should_ be in https://github.com/fermyon/developer/pull/266

karthik2804 (approved):

PR #268, opened 2023-02-01
exit child process on exit

rajatjindal:
This PR stops the background process started by the script.


tpmccallum:
Hi @rajatjindal 
Can you please explain this PR's code to me in terms of what it is designed to do/change and also the reason (good, bad or indifferent) for the build failing?
You can set me as a reviewer and we can also run through this together in Slack Huddle etc. in the coming days; if you like.
Thanks so much.
Tim

rajatjindal:
when we run the broken link checker script, we  start a spin process to start the documentation server. but when we exit, we don't explicitly close the spin server.

this PR  tries to fix that. The idea is that when the script is done, it will cleanup the child process and then exit.

and of course in the current version, it is not working correctly. (it exits with failure even when all process completely  successfully). I need to spend a little  more time to fix that.

tpmccallum:
Hi @rajatjindal 
Thank you for that great explanation. Looking forward to seeing a solution regarding exiting with failure even when all processes are completely successful. Please feel free to ping me (on Slack etc.) when you need me to return here and review/approve, or of you would like me to test anything from my end.
Thanks again
Tim

rajatjindal:
I think this is ready for review now. I tried many approaches, but only this one seems to work.

will be good to get review from @radu-matei @karthik2804 as well

karthik2804 (approved):
LGTM!

tpmccallum:
Thank you @rajatjindal and @karthik2804 
Great idea @rajatjindal (to keep an eye on it in the short term to make sure that there are no side effects)
Kind regards
Tim

PR #275, opened 2023-02-02
update 2023 event list

flynnduism:
Updates the carousel on the homepage with the next couple of forthcoming events:

events-carousel

Creating a follow-up issue to discuss a better way to publish/expire events as content objects going forward, rather than this messy, hard-coded approach.

---

Signed-off-by: flynnduism <ronan@fermyon.com>


tpmccallum (approved):
Looks great Ronan,
Just closed off the frontmatter and removed the comments from within the frontmatter (they were getting stuck in the linting as H1's

Issue #279, opened 2023-02-02
CLI reference for `spin` is only in the `cloud` section of the docs


mikkelhegn:
The `spin` cli reference is only available in the TOC of the Cloud section. I'd love for it to also be available in the Spin section of the docs.

tpmccallum:
Fixed by https://github.com/fermyon/developer/pull/285

We now have CLI reference resources (complete with appropriate template-based sidebar menu) in Spin and Cloud.
We now have doc contribution resources (complete with appropriate template-based sidebar menu) in Spin, Cloud and Bartholomew Docs

PR #284, opened 2023-02-03
Spin cli symlink

tpmccallum:
Signed-off-by: tpmccallum [tim.mccallum@fermyon.com](mailto:tim.mccallum@fermyon.com)
Added symlink from cloud/cli-reference.md to spin/cli-reference.md
Added new menu item in Spin section
In reference to https://github.com/fermyon/developer/issues/279

### DIFF
--- a/content/spin/cli-reference.md
+++ b/content/spin/cli-reference.md
@@ -1,0 +1,1 @@
+../cloud/cli-reference.md
--- a/templates/spin_sidebar.hbs
+++ b/templates/spin_sidebar.hbs
@@ -29,6 +29,13 @@
         </ul>
 
         <p class="menu-label">
+            References
+        </p>
+        <ul class="menu-list">
+            <li><a {{#if (active_content request.spin-path-info "/spin/cli-reference" )}} class="active" {{/if}} href="{{site.info.base_url}}/spin/cli-reference">CLI Reference</a></li>
+        </ul>
+
+        <p class="menu-label">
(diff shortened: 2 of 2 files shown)


tpmccallum:
Couple of things with this PR

1) The build is failing because the markdown is linking to the symlink'd spin/cli-reference.md file (which technically does not exist). We could just exclude this specific file from the broken link checker (like we do with some other URLs)

2) When testing locally (for the first time) the new menu item (as per the hbs file) is present.
Screenshot 2023-02-03 at 1 40 21 pm

However, when clicking on the menu item (which displays the page and correct Spin URL path etc.) the Cloud menu is actually what is being rendered as the menu now. See image below.

Screenshot 2023-02-03 at 1 45 08 pm

tpmccallum:
This is due to the fact that the markdown file that is being symlinked has the `template = "cloud_main"` in the frontmatter.
I am wondering if it is possible to have the `template` generated dynamically, based on the location of the `.md` file. For example, all files in the cloud folder adhere to the `cloud_main` or something.
The reason is, it would be great to also symlink other related docs such as "Contributing". The contents are the same (all pushing back to developer repo) but need to reside in the menus for Cloud, Spin and Bartholomew.

tpmccallum:
Had a good chat with Karthik on this one. The way in which the templates work will not allow us to use symlinks.
Closing this now in favour of another alternative approach.

PR #285, opened 2023-02-03
Implement common docs across different products

tpmccallum:
Signed-off-by: tpmccallum [tim.mccallum@fermyon.com](mailto:tim.mccallum@fermyon.com)

This PR deliberately produces the following duplicate contribution documentation pages, so that when a user is viewing via each of the different templates (`cloud_main`, `spin_main` or `bartholomew_main`) they have the opportunity to see and click on the Contribute to Docs menu item. Which are relevant to all three areas.

/bartholomew/contributing-docs
/cloud/contributing-docs
/spin/contributing-docs

This PR also deliberately produces the following duplicate cli reference documentation pages so that when a user is viewing via each of the different templates (`cloud_main`, or `spin_main`) they have the opportunity to see and click on the CLI Reference menu item. Which are relevant to both areas.

/spin/cli-reference
/cloud/cli-reference

This PR addresses this issue https://github.com/fermyon/developer/issues/279

**Making this future proof**

@karthik2804 is going to create a CI automation which will test to see that the above duplicate docs share duplicate content (upon build/commit). This CI script will raise an error that will prompt the contributor to update the other documents (which have outdated content). This CI script will ensure that no content will ship out of sync.

@karthik2804 please note that the only line in the `.md` file that will be different is the `template = x_main` (around line 2 or 3). This is obviously due to the fact that each file is designed to be rendered to a separate part of the site (via templates).

**Why this is being done, this way**
An earlier attempt at using symlinks was unsuccessful ( see https://github.com/fermyon/developer/pull/284 ). Symlinks that share a file's contents also share that files `template` and this results in incorrect rendering of the page.


tpmccallum:
Checks naturally failing at this point because the new files being added here are broken links until this PR is merged (and the files actually exist). We can ignore these broken links to merge.

vdice (commented):
First of all, these docs are some of the most thorough, detailed and helpful I've seen @tpmccallum 👏 !  I myself have learned quite a bit, especially with regards to markdown :)

Appreciate the context in the original description re: the reason for this approach.  I did want to broach the possibility of another approach for the `contributing-docs.md`: I wonder if we might place just one copy in either a new 'Developer' section (that would concern items relevant to this Developer website or otherwise more general than the projects (Spin, Cloud, Bart) collected here) or somewhere discoverable at a remove from the project tabs.  Then, no extra CI/checks would be necessary for syncing mutliple files.  I'd assume we could still place links to this doc under each project -- they'd just all redirect to this canonical page.

Also, for the Spin CLI reference -- have we looked into tools/approaches to auto-generate documentation from a known spin binary or version?  This happens to be top of mind as there is currently a PR in fermyon/spin updating its cli scaffolding tool (`clap`) to the next major version, which will require updating this reference anew (all output will change).  Beyond this major event, it'd be great to find an automated approach so that the common cases of updated flags/args/cmds don't require manual update efforts to this doc.  The bonus being, such automation should hopefully cover keeping multiple versions in sync.

karthik2804:
I agree with @vdice about moving the contribution docs to a common location. We may still need the ci test for sync across duplicates for the cli reference.

itowlson:
@vdice Naming nitpick: Not sure if you're proposing "Developer" as an internal organising folder in this repo or as a visible category on the Web site, but I have reservations about it as a name, especially if it's that latter.  This is ALL developer docs: I worry that developers will come here looking for info about _using_ Spin and Cloud, see "Developer", go THAT'S ME, and get frustrated that it's not relevant to them at all.  Maybe "contributor/contributing"?

tpmccallum:
Hi

**TL;DR** 
@vdice I love the idea of a canonical page, i.e. a single source of truth without CI redundancy checks etc.
@itowlson's point about making the docs unambiguous is also spot on. 
Completely on board.
As a phase 1, would you all mind merging this PR as is "for now"? The reason being it is an improvement over and above what we currently have (users don't see a CLI reference on the Spin menu; which is technically where the reference should be, and it would also be great to have a pathway for users to contribute in the immediate future). I can then add an issue that addresses how we could set up a central area containing canonical pages.

---

The phase 2 part of this work could present the content to the user in exactly the same way (as this PR does) with the added bonus that we have restructured the layout of files to avoid redundancy and the need for CI checks.

itowlson:
Totally happy for this as an interim solution and apologies for inadvertent bikeshed

vdice (approved):
Sounds good @tpmccallum; LGTM on these great additions!

vdice:
Oops, a merge conflict crept in; once addressed, feel free to merge when ready.  The event will then trigger an auto-deploy to https://developer.fermyon.com.

tpmccallum:
This issue https://github.com/fermyon/developer/issues/291 is the discussion for phase 2 of the common docs solution.

PR #293, opened 2023-02-08
chore(deploy.yml): only attempt website deploy if org is fermyon

vdice:
Only deploy website if org is fermyon (disable for forks)


itowlson (approved):

vdice:
(build failure to be addressed in https://github.com/fermyon/developer/pull/294)

PR #308, opened 2023-02-09
Removing redundant files, because redirects are in place and working perfectly

tpmccallum:

Removing redundant (legacy) files to ensure no unnecessary edits can occur. 
We have implemented a single-source-of-truth approach where any single file intended to be shared among different projects, i.e. `contributing-docs`, is disseminated via the `common` folder.

The redirect rules are applied when Spin starts; adjusting routes which are pre-configured in the `spin.toml` so there is no need to keep legacy files around, once redirects are in place.

Screenshot 2023-02-09 at 11 21 53 am


tpmccallum:
This has been tested locally and works perfectly.

itowlson (approved):

PR #315, opened 2023-02-09
Create dropdown to list projects

karthik2804:
Creates a dropdown navbar to switch between projects.

image


tpmccallum:
This is really cool @karthik2804 

Just a quick observation Re: hand-held/mobile devices.

The initial rendering is fine i.e. clicking on the three-horizontal-bars in the top-right corner correctly renders the project options.

image

image

However, once the user clicks on a specific project i.e. Spin, they can not access that original dropdown to list the projects (as per the initial page load).

image

image

tpmccallum:
Ah, when comparing this new PR to how the original developer docs render on mobile devices I can see that the aforementioned issue is not actually addressed by this PR. This PR improves the desktop version of the site (by providing a dropdown) but the mobile (smaller screen resolutions still just retain the older/original design.

Apologies for the confusion.

This has brought up a good point though; is it possible to add the same type of project menu to the header on all smaller screen sizes? mobile device? i.e.

image

flynnduism (approved):
This works really nicely @karthik2804 /lgtm!

We should use this same approach on fermyon.com too

PR #330, opened 2023-02-14
[WIP] add deploy/cleanup preview

rajatjindal:

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -1,0 +1,43 @@
+name: Deploy Preview
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened]
+
+jobs:
+  cloud-login:
+    uses: "fermyon/actions/.github/workflows/auth.yml@main"
+    secrets:
+      gh_username: ${{ secrets.DEV_DOCS_PREVIEW_GH_USERNAME }}
+      gh_password: ${{ secrets.DEV_DOCS_PREVIEW_GH_PASSWORD }}
+      gh_totp_secret: ${{ secrets.DEV_DOCS_PREVIEW_GH_TOTP_SECRET }}
+    with:
+      fermyon_deployment_env: developer-docs-preview
+
+  deploy-preview:
+    runs-on: "ubuntu-latest"
+    needs: [cloud-login]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Retrieve saved cloud token
+        uses: actions/download-artifact@v3
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,43 @@
...
+
+  cleanup: 
+    runs-on: ubuntu-latest
+    needs: [deploy-preview]
+    if: always()
+    steps:
+    - uses: geekyeggo/delete-artifact@v2
+      if: always()
+      with:
+        name: developer-docs-preview.json
```
radu-matei, line 43: This is feedback for the upstream action and not this use case in particular, but: I think it's really easy to miss adding the cleanup step and accidentally leave a token as an artifact.

I'd rather the action generate the token again using the secret rather than reusing it.

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,43 @@
...
+      - name: Deploy preview on cloud
+        uses: rajatjindal/fermyon-preview-deploy@main
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          fermyon_token_file_name: "developer-docs-preview.json"
+
+  cleanup: 
+    runs-on: ubuntu-latest
+    needs: [deploy-preview]
+    if: always()
```
radu-matei, line 38: When does the `if: always()` condition run?

.github/workflows/deploy-preview.yml
```diff
@@ -0,0 +1,43 @@
+name: Deploy Preview
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened]
+
+jobs:
+  cloud-login:
+    uses: "fermyon/actions/.github/workflows/auth.yml@main"
```
radu-matei, line 9: Same comment re: the action persisting the token to an artifact.

.github/workflows/deploy-preview.yml
rajatjindal: yeah, I need to find a better story on how we can pass along secrets to forks safely.

.github/workflows/deploy-preview.yml
rajatjindal: its run at the end of workflow and runs even if other steps/jobs have failed in the workflow

rajatjindal:
I am going to close this PR in favor of #332

.github/workflows/deploy-preview.yml
radu-matei: Ah, got it, thanks!

PR #332, opened 2023-02-14
Trying out preview deployment from fermyon org

rajatjindal:
this is the PR that will add preview support for developer docs. Few gotchas:

- It can deploy previews only if PR is from a branch in `fermyon` org only. This is because of security considerations.
- we can deploy a max of 5 preview only right now.
- The action can delete old previews when it is out of limits (disabled right now), we should probably enable it. Pls let me know
- The action deletes the preview when the PR is closed. (merged or not).
- It currently uses browser automation to do the cloud login
- It is currently using a Github account I have setup using one of my personal domain (I am working on configuring a Fermyon account for this. will update the secrets in repo once we are ready for that.)

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -1,0 +1,25 @@
+name: deploy docs preview
+on:
+  pull_request:
+    branches: "main"
+    types: ['opened', 'synchronize', 'reopened', 'closed']
+
+jobs:
+  deploy-preview:
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'fermyon' }}
+    name: Build and deploy
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Setup `spin`
+        uses: fermyon/actions/spin/setup@v1
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: build and deploy preview
+        uses: fermyon/actions/spin/preview@v1
+        with:
+          fermyon_token: ${{ secrets.FERMYON_CLOUD_TOKEN }}
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          undeploy: ${{ github.event.pull_request && github.event.action == 'closed' }}


rajatjindal:
capturing the feedback from Radu on the other PR that I am closing in favor of this repo:

> This is feedback for the upstream action and not this use case in particular, but: I think it's really easy to miss adding the cleanup step and accidentally leave a token as an artifact.

> I'd rather the action generate the token again using the secret rather than reusing it.

.github/workflows/cleanup-preview.yml
```diff
@@ -0,0 +1,44 @@
+name: Cleanup Preview
+on:
+  pull_request:
+    branches: [main]
+    types: [closed]
+
+jobs:
+  cloud-login:
+    if: ${{ github.repository_owner == 'fermyon' }}
+    uses: "fermyon/actions/.github/workflows/auth.yml@main"
```
vdice, line 10: nit: should we eventually pin the action versions (here and `rajatjindal/fermyon-preview-deploy` below) to tags as opposed to `main`?

vdice (commented):
This is an exciting and helpful addition!  What exists now seems like a great foundation to build upon.  LGTM; I'll also let the other requested reviewers chime in...

> The action can delete old previews when it is out of limits (disabled right now), we should probably enable it. Pls let me know

I'd perhaps vote for keeping it disabled for now, at least until the upstream action supports notifying (via GH comment) when a preview is deleted in combination with supporting redploys via PR author comments.

However, if left disabled, what does the action do when the account app limit is reached?  Is a GH comment added to PRs where preview deploys aren't possible?

Either way, both situations will be improved by the possibility of increasing the app limit for the configured deploy preview account, so hopefully we can prioritize this.

tpmccallum:
This deploys a version of the developer docs onto Fermyon Cloud so a contributor can review their changes.

At the moment, there is a pending topic: how the actions will be run, i.e. where does the action live? Should it be an official Fermyon action or a PoC on a user's GitHub account?

The personal access token PR < https://github.com/fermyon/spin/pull/1201 > can proceed to a point where the personal access token can be the authentication mechanism for a particular user. However, this is still a work in progress.

tpmccallum (approved):

PR #339, opened 2023-02-16
Homepage: consider not opening new tabs

tpmccallum:


itowlson (approved):
I'd be delighted to see this merged but it probably needs to be socialised around the folks who originally wanted it the current way...

PR #357, opened 2023-02-20
Update extending-and-embedding.md

tpmccallum:


itowlson (approved):
To be clear, this is a quick fix to remove a broken link to the former `publish` crate, and it might be good if the commit message reflected that.  A true 'update' would need to be broader than this to bring it into line with how Spin currently works.  But within that narrow purpose it looks good!

tpmccallum:
Thanks, Ivan; I agree.

radu-matei:
You could point to https://github.com/fermyon/spin/pull/1166 in the meantime.

PR #377, opened 2023-02-28
Http triggers using https

tpmccallum:

Screenshot 2023-02-28 at 11 01 57 am

Screenshot 2023-02-28 at 11 05 10 am


content/spin/https-trigger.md
```diff
@@ -0,0 +1,55 @@
...
+
+---
+- [HTTP vs HTTPS](#http-vs-https)
+- [Routing](#routing)
+- [Defining the HTTP Object](#defining-the-http-object)
+
+## HTTP vs HTTPS
+
+The [Spin HTTP trigger document](./http-trigger.md) provides a great overview of the HTTP trigger, as well as implementation details around the WebAssembly component model and how it
+is used in Spin. However, there may be certain situations where you would want to accept only HTTPS requests (and reject HTTP requests). This raises questions like, do I have to provide a TLS certificate, and so forth?
```
itowlson, line 14: It's not clear to me why we don't answer those questions in the HTTP article - why does the user need to discover and navigate to a separate page?

rajatjindal:
I am checking why links still show up as broken

content/spin/https-trigger.md
```diff
@@ -0,0 +1,55 @@
...
+export SPIN_TLS_CERT=publickey.pem
+```
+
+<!-- @nocpy -->
+
+```bash
+export SPIN_TLS_KEY=pkcs8.key
+```
+
+Once created, both of the above environmental variables can be used as arguments of the `spin up` command.
```
itowlson, line 55: This is a bit confusing - do you mean I have to pass these to `spin up` as command line options?  And if so how?  (The answer is no, if they are set when you run `spin up` then Spin will pick them up automatically.  You can pass by arguments _instead_ using the `--tls-*` command line options.)

itowlson (changes requested):
I feel this would be easier to find, and would result in less content that is tangential to the key point, as a section in the main HTTP page.  A lot of this page, I'm afraid, feels like it's padded to justify being a separate page, when all it really needs to say is "use these two environment variables / CLI flags."  Sorry.

tpmccallum:
Thanks @itowlson 
Instead of this new page, we can create a succinct subsection on the existing HTTP Trigger page. That will be much neater and straight to the point. I will close this one out.
Thanks 
Tim

PR #384, opened 2023-02-28
Unify spin.toml in Spin key-vaule docs

ThorstenHans:
`spin.toml` had quite some differences across Rust, TypeScript, and TinyGo. With this commit, all manifests rely on the same naming-pattern, provide a description, and have inline objects formatted the same way

## Note on `bart check`

`bart check` ran without errors / warnings for the file I touched. However, there are some findings in other files located in `/content/spin`


content/spin/kv-store.md
```diff
@@ -65,7 +65,7 @@ As previously documented, you can go ahead and [create a new Spin application fr
 <!-- @selectiveCpy -->
 
 ```bash
-$ spin new http-rust spin-key-value
+$ spin new http-rust spin-rust-key-value
```
itowlson, line 68: Given that all these blocks are under per-language tabs, I'm not sure if there's any value in having different names.  Would it be bad if we called them all plain `spin-key-value`?  I don't have strong views, more raising the question for @tpmccallum.

itowlson (commented):
Thanks for this!  I had one thought but probably as much for us as for you...!

itowlson:
I think the failing action may be a link already in `main` that is "in waiting" but not live yet.

ThorstenHans:
> Thanks for this!  I had one thought but probably as much for us as for you...!

Thought about that too. Will update the PR once back at the mac

itowlson (approved):
LGTM - thanks!

content/spin/kv-store.md
tpmccallum: Consistency is welcome across all tabs; greatly appreciated and about to be merged ...

tpmccallum:
Thanks again @ThorstenHans 
Greatly appreciated!

PR #385, opened 2023-02-28
Explain the need of the http crate for outbound HTTP requests with Rust

ThorstenHans:
If developers create the app using templates other than http-rust, they must ensure that the http crate is listed as dependency in `Cargo.toml`. Otherwise APIs like `http::Request::builder()` are not available.


content/spin/rust-components.md
```diff
@@ -102,6 +102,8 @@ fn hello_world(_req: Request) -> Result<Response> {
 }
 ```
 
+In the previous snippet, we used `http::Request::builder()`, which is provided by the `http` crate. Ensure that `http` is part of your dependencies in `Cargo.toml`. You must manually add the dependency for projects created using templates other than `http-rust`.
```
itowlson, line 105: Suggestion: it might be good to indent this paragraph as a "side note", similar to the way the hosts remark is indented at line 147, so as to avoid breaking the flow for people who are using the template (which we currently expect to be the most common case).

itowlson, line 105: Another suggestion for the flow of this paragraph, something along the lines of:

> The `http::Request::builder()` method is provided by the Rust `http` crate.  The `http` crate is already added to projects using the Spin `http-rust` template.  If you create a project without using the template, you'll need to add the `http` crate yourself via `cargo add http`.

This flow puts "whether you need or don't need to worry about this" a bit earlier, rather than the kind of unqualified warning to "Ensure...", if that makes sense?

ThorstenHans: LGTM, however if went for `without using this template` instead of `without using the template`

itowlson: Oh, good call - thanks for the improvement!

itowlson (approved):

PR #396, opened 2023-03-01
Hide log spew in `npm run spin`

itowlson:
In recent versions of Spin, guest output is printed to the console by default.  Unfortunately, something in either `nodemon` or Bartholomew is causing the index page to be re-requested every second or less, and this now prints a _lot_ of diagnostic data every single time.

This PR passes `--quiet` to `spin up` so the noise goes away, though it might be useful to know why the page is re-requested so often in the first place.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "yaml-front-matter": "^4.1.1"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


karthik2804 (approved):
Thanks!

karthik2804:
#397 fixes CI.

PR #398, opened 2023-03-01
Explain that npm run styles is designed to be a background process

tpmccallum:

### DIFF
--- a/content/common/contributing-docs.md
+++ b/content/common/contributing-docs.md
@@ -223,7 +223,7 @@ npm run build-index
 
 ### 6.3 How To Properly Edit CSS Styles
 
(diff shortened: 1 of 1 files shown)


karthik2804:
It would also be nice to mention that it is not necessary unless editing the `scss` file as the `css` files are already checked in.

itowlson (changes requested):
It's great to have this clarified - thanks!  However, my puzzlement was related more to the README - we should update that to reflect this (or remove the instruction to `npm run styles` altogether, as that's not needed for a casual "run it locally").  Most interested parties - at least in this sample of one overconfident kiwi - are going to at least start by following the readme rather than working through the contributing guide.

To be clear, the content here is great, the request changes is to have it on the readme too.  Thanks!

itowlson (approved):
My worry about the readme is covered off in #399 - thanks!

PR #409, opened 2023-03-02
run docs tests before npm test

karthik2804:


tpmccallum (approved):
I see this is just repositioning and therefore don't see any issues approving.

Would it be a good idea to create a new PR that uses Bart 0.6 instead of 0.3.0?

karthik2804:
Yea can add that as a separate PR also adding support for the shortcodes.

PR #415, opened 2023-03-02
fix inline codeblock styling inside table

### LINKED ISSUE #414: Indent on first line of inline monospace block
I have some inline monospaced elements, created in Markdown using single backtick and no embedded leading space.  For example:

image

When I render them on the docs site, they appear to have a leading space or small leading indent:

image

Expected behaviour is that the first characters of each line would align vertically.

karthik2804:
closes #414 

image
image
image


itowlson (approved):
Live footage of Ivan approving CSS PRs

image

PR #419, opened 2023-03-05
Update Spin version in OCI documentation

tpmccallum:


content/spin/spin-oci.md
```diff
@@ -141,6 +161,16 @@ Lastly, let's run this Spin application.
 
 {{ blockEnd }}
 
+{{ startTab "v0.10.0"}}
+
+<!-- @selectiveCpy -->
+
+ ```bash
+ spin up --from-registry ghcr.io/USERNAME/spin-react-fullstack:v1
```
itowlson, line 169: Consider using `-f` here.  You should only need to use `--from-registry` if Spin can't recognise the string as a registry reference.  (Please tell me if this isn't working for your test case!)

tpmccallum: I have the `-f` for the `0.10.0` version (just below that `0.9.0` tab syntax).
Wes `-f` `--from` implemented in `0.9.0`? If so happy to update as per this review commend; otherwise the `0.10.0` tab is useful to show the difference between running `.9` and `.10` just in case a user is on an older version and following these docs.

tpmccallum: Screenshot 2023-03-06 at 9 47 28 am

karthik2804: I'd potentially put v0.10.0 before v0.9.0 ad also if there is no change we should possibly just make the blocks `>=v0.9.0`, just a thought.

tpmccallum: Sure, that's a great point @karthik2804 
Let me re-org and ping back here when ready.
Gimme a sec ...

itowlson: @tpmccallum No, you are correct, `-f` is 0.10 onwards.  Apologies, I didn't open the surrounding lines and misread this as new 0.10 content - sorry!  _disappears into hedge_

itowlson: @karthik2804 I quite like the explicitness of spelling out individual versions, and I think it plays nicely with sticky tabs?  So if I select 0.9 then I get 0.9 _everywhere_ rather than needing to select `>= 0.9` / `<= 0.9` / `0.8 though 0.11` etc. as I move through different parts of the docs.

itowlson: +1 to latest version first though

tpmccallum: Ok, that looks great; we should make that a convention i.e. the latest version from left to right in the multi-version code block.
Screenshot 2023-03-06 at 10 00 39 am

tpmccallum: Also, having more than two versions (like before `0.8.0`, `0.9.0`, & `0.10.0`) was probably unnecessary and too busy for the eye. Great feedback on the `>=`

tpmccallum: Thanks @itowlson 
This is a good feedback hybrid; let me revise it slightly, and I think we might have it nailed. Another sec ...

content/spin/install.md
```diff
@@ -90,7 +90,7 @@ To install the canary version of spin, you should pass the argument `-v canary`.
 
 {{ startTab "Windows"}}
 
-On Windows systems you can download <a href="https://github.com/fermyon/spin/releases/tag/v0.9.0" class="spin-install" id="spin-install-windows">the Windows binary release of Spin</a>.
+On Windows systems you can download <a href="https://github.com/fermyon/spin/releases/tag/v0.10.0" class="spin-install" id="spin-install-windows">the Windows binary release of Spin</a>.
```
itowlson, line 93: Have we considered changing this to `latest` and Never Speaking Of It Again?

content/spin/kv-store.md
```diff
@@ -23,7 +23,7 @@ enable_shortcodes = true
...
+Spin applications are best suited for event-driven, stateless workloads that have low-latency requirements. Keeping track of the application's state (storing information) is an integral part of any useful product or service. For example, users (and the business) will expect to store and load data/information at all times; during an application’s execution. From [Spin v0.9.0](https://www.fermyon.com/blog/spin-v09), onwards we've added support for key/value storage. This is a storage option for applications that need data in the form of key/value pairs and are satisfied by a BASE consistency model. Workload examples include general value caching, session caching, counters, and serialized application state.
```
itowlson, line 26: At this point maybe remove the version qualification?  We are not really making an effort to keep the docs relevant to 0.8 and earlier, so maybe treat K/V as just a feature that we have.

content/spin/kv-store.md
```diff
@@ -39,13 +39,13 @@ In all of these previous examples, to persist data within your application, a se
...
 
-In the future, it may be possible to embed other database technologies into Spin (as v0.9.0 has shown here with SQLite). For example, future Spin SDK updates may target other databases such as Redis; which may allow you to build applications using Redis via minimal Spin configuration without the need to locally install and maintain your own Redis instance.
+In the future, it may be possible to embed other database technologies into Spin (as v0.9.0 onwards has shown here with SQLite). For example, future Spin SDK updates may target other databases such as Redis; which may allow you to build applications using Redis via minimal Spin configuration without the need to locally install and maintain your own Redis instance.
```
itowlson, line 42: The parenthesis here doesn't really add anything - could say `... into Spin, in the same way as SQLite.` if you want to keep something, but I'd drop "which version showed it".

content/spin/kv-store.md
```diff
@@ -39,13 +39,13 @@ In all of these previous examples, to persist data within your application, a se
...
 
 Let's get started with creating and deploying your first Spin application that uses this new key/value storage mechanism.
 
 ## Tutorial Prerequisites
 
-First, follow [this guide](./install.md) to install Spin v0.9.0 or later. To ensure you have the correct version, you can check with this command:
+First, follow [this guide](./install.md) to install Spin v0.9.0 **or later**. To ensure you have the correct version, you can check with this command:
```
itowlson, line 48: Just "the latest" or "a recent version of"?  You can put a note after the `spin --version` block to say "If the version is 0.8 or earlier, you'll need to upgrade to a more recent version."

content/spin/kv-store.md
```diff
@@ -186,13 +186,13 @@ In this section, we use the Spin SDK to open and persist our application's data
 
 ### The Spin SDK Version
 
-If you have an existing application and want to try out this new key/value feature (i.e. a previous application made using the `http-rust` template). You might need to update your existing application's `Cargo.toml` to point to v0.9.0 of the Spin SDK, as shown below:
+If you have an existing application and want to try out this new key/value feature (i.e. a previous application made using the `http-rust` template). You might need to update your existing application's `Cargo.toml` to point to v0.10.0 of the Spin SDK, as shown below:
```
itowlson, line 189: This paragraph now becomes a bit fiddly, because if they're on 0.9 they _don't_ need to upgrade.  Maybe: If you have an existing app ... check the Spin SDK reference in Cargo.toml.  If it refers to version 0.8 or earlier, update it to match your upgraded version of Spin.  Or something like that?

I'd also consider dropping "new" when mentioning the K/V feature, but maybe we're not at that point yet?

content/spin/kv-store.md
```diff
@@ -543,8 +543,8 @@ You can read the [improvement proposal for key/value support](https://github.com
 {
   "@context": "https://schema.org",
   "@type": "VideoObject",
-  "name": "Key/Value Store preview - Spin 0.9.0",
-  "description": "Overview of the key/value store in Spin v0.9.0",
+  "name": "Key/Value Store preview - Spin 0.10.0",
```
itowlson, line 546: Is this still a preview?  Is it specific to 0.10?  I'd suggest dropping the version from this point on.  Generally this page (with references to the improvement proposal etc. and discussion of motivation and future direction) feels like a bit "moment in time meet the new thing" when it maybe needs to shift to "there is a thing, here is how to use it."

But that may be a larger change than you want to make in this PR.

content/spin/install.md
tpmccallum: `latest` it is i.e. `https://github.com/fermyon/spin/releases/latest` works like a charm.

content/spin/kv-store.md
tpmccallum: Done

content/spin/kv-store.md
tpmccallum: Parenthesis was removed, and it reads much better, thanks.

content/spin/kv-store.md
tpmccallum: Good point; the preview text was removed.

content/spin/kv-store.md
tpmccallum: Again, great point, changed in new commit.

content/spin/kv-store.md
tpmccallum: This has been reworded now, thanks.

content/spin/spin-oci.md
tpmccallum: Done

content/spin/kv-store.md
```diff
@@ -23,7 +23,7 @@ enable_shortcodes = true
...
+Spin applications are best suited for event-driven, stateless workloads that have low-latency requirements. Keeping track of the application's state (storing information) is an integral part of any useful product or service. For example, users (and the business) will expect to store and load data/information at all times; during an application’s execution. [Spin](https://www.fermyon.com/blog/spin-v09) has support for applications that need data in the form of key/value pairs and are satisfied by a BASE consistency model. Workload examples include general value caching, session caching, counters, and serialized application state.
```
itowlson, line 26: Nit: The 0.9 blog post link is a bit peculiar as the link for the unvarnished "Spin" text.  You could extend the link to "Spin has support."  Or you could just not worry about it, that would be fine too!

itowlson (approved):
LGTM, thanks for all the fixes and all the attention to detail!

Issue #369, opened 2023-02-24
CI/deploy for search index (data.json)


vdice:
@karthik2804 and I were discussing the following pain point: Currently, this website's search index (`static/data.json`) needs to be updated in each applicable PR (any PRs updating documentation content, no matter how small, are applicable).  There is a command to do so (`npm run build-index`) and a note/reminder on the PR template, but it is still an extra step for contributors -- not to mention it introduces the common scenario of merge conflicts occuring to in-flight PRs as others are merged in.

Therefore, we'd like to find a solution where CI/automation does this chore work in the background, removing the need for contributor PRs to update this file (and eliminating the frequent merge conflict potential).

In brainstorming with @karthik2804, these are the top candidates that seemed preferable to some other scenarios discussed.

### Context

Currently, the website itself is published to bindle as a Spin app after cloning this repo at a configured branch and/or commit.  The live website consists of a `spin up --bindle-id` command using the published bindle ID.

### Publish/deploy from OCI registry reference

Added [below](https://github.com/fermyon/developer/issues/369#issuecomment-1444658594), bumped up to top potential candidate (bonus: dogfoods Spin's newly-added OCI reference support):

On every push to `main`, an automation agent (eg GH action) would clone the repo and run `npm run build-index`, then publish the website (Spin app) to an OCI registry (obvious choice: `ghcr.io/fermyon/developer`) via `spin registry push`.  We'd remove Bindle from the equation and the [website Nomad job](https://github.com/fermyon/developer/blob/main/deploy/fermyon-developer.nomad) would be updated to run from the pushed image reference, eg `spin up -f ghcr.io/fermyon/developer:<tag>`.  As the search index is commited in the form of an image, we'd `.gitignore` the file.

Considerations: To achieve parity with current deployment workflows, which allow for deploying from arbitrary branches/commits on the fermyon origin, we'd perhaps want to add publishing of images on PRs (again just for branches on the fermyon org).  Or, we could decide that ad hoc/out-of-band publishing by those with access to this repo's container registry is sufficient prior to non-main deploys.  For the default website image, we could use a mutable tag like `latest` that would be updated on every push to main.  Alternatively, we could follow the convention of using unique/"immutable" tags but would want to look into storage costs and/or automated cleanup.

Alternative:

### Deploy website from branch other than `main`

This retains current behavior of tracking the search index in git, albeit via a different branch from `main`.

[...]

PR #423, opened 2023-03-06
ci(*): update publish/deploy to use ghcr.io; untrack static/data.json

vdice:
Updates publish/deploy logic to use an OCI registry (first choice: ghcr.io).  In doing so, we generate the search index (`static/data.json`) at time of publishing, removing the need to track in version control.  (Ref https://github.com/fermyon/developer/issues/369)

Placing in Draft for now as there is one blocker:  Getting errors attempting to push this Developer website Spin app to ghcr.io (see https://github.com/vdice/developer/actions/runs/4326339883/jobs/7553631989).  An auth error is produced which is odd/unexpected as I can publish other Spin apps to ghcr.io with the same flow.  Additionally, this Developer site can be published to docker.io/DockerHub.  Need to investigate.

_edit: Update on the blocker.  As mentioned, it only occurs with ghcr.io due to its auth token mechanism, which appears to only be valid for 1 minute and is used for all layer push requests.  This, in tandem with spin's sequential image layer pushes (and pulls) causes the timeout.  We are tracking updating spin to push and pull concurrently, which will benefit regardless of registry (see https://github.com/fermyon/spin/pull/1296) but should also unblock pushing to ghcr.io._


bacongobbler:
Does the action have write permissions to the package API? IIRC you need to add the following to your job metadata:

```
jobs:
  publish:
    runs-on: ubuntu-latest
    permissions:
      contents: read
      packages: write
```

https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages

karthik2804:
@bacongobbler I could reproduce this issue from my local CLI as well. I am able to push other packages but this one fails. So I am not entirely sure that it is a permissions issue.

itowlson:
I see the same thing as @karthik2804 - from my local machine:

* I can push "hello world" applications to ghcr
* I can push the developer docs site to my localhost registry
* But pushing the developer docs to ghcr fails with 401 Unauthorized

I did wonder if it could be a size issue, as it takes a long time to upload, but the developer docs package looks to be about 35MB, which doesn't seem unreasonable.  And of course it would be very rude to report that as "authentication required" anyway.

@radu-matei, I think you probably know more about OCI than the rest of us - any insight into what might cause this behaviour and response, or how to diagnose it?

vdice:
Superseded by https://github.com/fermyon/developer/pull/588

PR #428, opened 2023-03-07
css adjustments for docs

flynnduism:
* top bar: reduces/delays the blur effect on the projects menu
* scales down table font-size, padding and tweaks styles
* fix the left-side indentation on code blocks


karthik2804:
I see the `data.json` is checked in. Can we remove that.

flynnduism:
@karthik2804 should I `.gitignore` that?

karthik2804:
Nope, it is required for the search index until we get the CI for it sorted out. Just not commit it with this PR.

static/sass/styles.scss
```diff
@@ -654,22 +680,25 @@ html.dark-theme {
...
-					display: none;
+					// display: none;
 					position: fixed;
-					left: 0;
+					left: 500%;
 					top: 0;
 					width: 100vw;
 					height: 100vw;
+					opacity: 0;
+					transition: backdrop-filter ease-in-out 1s 0.1s;
```
karthik2804, line 694:
```suggestion
```

static/sass/styles.scss
```diff
@@ -654,22 +680,25 @@ html.dark-theme {
...
+					opacity: 0;
+					transition: backdrop-filter ease-in-out 1s 0.1s; 
 				}
 
 				&:hover {
 					+.overlay {
-						display: block;
+						// display: block;
+						left: 0;
+						opacity: 1;
```
karthik2804, line 701:
```suggestion
						left: 0;
						opacity: 1;
						transition: backdrop-filter ease-in-out 1s 0.1s; 
```

karthik2804 (commented):
Just a small tweak to the CSS

itowlson:
This helps enormously - thank you!  Here is what the worst of the tables looks like with this change:

image

From looking at the sites we shared, I wonder if a possible further improvement would be to reduce the left margin.  I realise we don't want long lines so this would mean increasing the right margin, but that would mean we could show the page TOC at lower window sizes and would have more flexibility to extend tables rightward if need be.  Both Hashi and MS Learn have a slightly wider site TOC, bring their bodies much closer in to that TOC, and leave most of their whitespace to the right (with the page TOC if they have one).  We don't need or want to ape them, of course, but do you think something like that might work for us too...?

Regardless of further tweaks - I will take this in a heartbeat - much appreciated.

itowlson (approved):

karthik2804:
@itowlson, do you mean something like this? 

image

itowlson:
@karthik2804  Let me share the comparisons I shared with @flynnduism...

Hashi:

image

Microsoft:

image

I feel like the bodies of those are maybe more central than they are in your mockup?  Perhaps assisted by the wider (nested) site TOC?  I am not sure though.  I agree we would not want it too lopsided to the left!

PR #434, opened 2023-03-09
fix(spin docs): /spin/configuration no longer valid

vdice:
- Noticed that https://developer.fermyon.com/spin/configuration is no longer a valid path, so updated the one redirect rule it was associated with.

- Also made updates to the `spin_sidebar.hbs` file -- but not confident they are truly necessary or if the current `request.spin-path-info`values were intended?  (pls advise on this one.)


itowlson (approved):
Whoa, thanks for the spot.  Sorry about that.  I wonder why the link checker didn't pick these up.

Do you think we should add a redirect from the old `configuration` route to the new `writing-apps` one as well?

karthik2804:
@itowlson - the link checker does not look at the redirects. it just looks at the links it finds through the content.

tpmccallum:
I love how the sitemap is generated upon request also.
I see that these new changes are all correct in the sitemap without any extra work required
https://developer.fermyon.com/sitemap.xml

PR #440, opened 2023-03-09
ci(*): add test-website.sh; use in build.yml

vdice:
I noticed we didn't appear to have a quick 'spin up' test for this website to verify it runs as expected (and, minimally, returns 200).  It can be handy to double-check this on PRs, so added one.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,3 +40,7 @@ jobs:
       - name: Run npm tests
         run: |
           npm test
+
+      - name: Test website
+        run: |
+          npm run test-website
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "bundle-scripts": " parcel build static/js/src/main.js --dist-dir static/js --no-source-maps",
     "styles": "parcel watch static/sass/styles.scss --dist-dir static/css",
     "test": "npx markdownlint-cli2 content/**/*.md \"#node_modules\" && npm run check-broken-links",
+    "test-website": "./tests/test-website.sh",
     "build-index": "node md_parser.mjs --dir=./content/ --out=./static/data.json",
     "check-broken-links": ".build/check-broken-links.sh"
   }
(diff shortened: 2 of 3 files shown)


karthik2804:
Just a quick question - The `spin up` from the link checking should achieve what this one does right? or am I missing something?

https://github.com/fermyon/developer/blob/e50cd1d2c86cbfabc50faa29ad7bcc03bb79ebbc/.build/check-broken-links.sh#L53

vdice:
Oh, thanks @karthik2804.  I didn't think of the link-checker :)

It does technically catch the site not running correclty but it may be non-obvious... for instance when I introduce a typo into the spin.toml, I see the following (site doesn't come up within the 60s timeout, test exits):
```
$ npm run check-broken-links

> developer-fermyon-com@0.1.0 check-broken-links
> .build/check-broken-links.sh

starting developer portal
stopping process for 7756
cleanup_with_child 7756
stopping child 7821 (parent: 7756)
cleanup_with_child 7821
stopping child 7901 (parent: 7821)
cleanup_with_child 7901
stopping self 7901
stopping self 7821

$ echo $?
124
```

But I'd be down to close this PR and work with that script instead to perhaps add logic to produce clearer error(s) when the site just isn't running right.  Then we'd maybe want to update the name of the script from check-broken-links to something more general... WDYT?

karthik2804:
@vdice, I do not have a strong particular preference for either approach. We could make the `check-broken-links` more general - I will be adding more link-checking stuff into it next week. I am happy to approve this PR as well this may be better for cleaner separation of the functionality of scripts.

karthik2804 (approved):

vdice:
> On the other hand, the changes from this PR are small enough for it to be introduced into the check-broken-links, and naming it to something more general might be a good idea in any case.

Yea, I agree.  At this stage, I think this would be the cleanest approach for now.  I'll close this one and re-open with these changes.

PR #441, opened 2023-03-09
Change absolute glob pattern resolution to stay relative

kate-goldenring:
Small nit. Files are not converted to an absolute path rather stay relative

### DIFF
--- a/content/spin/writing-apps.md
+++ b/content/spin/writing-apps.md
@@ -263,7 +263,7 @@ files = [ "images/**/*.jpg", { source = "styles/dist", destination = "/styles" }
 
 The `files` field is an array listing the files, patterns and directories you want to include. Each element of the array can be:
 
-* A glob pattern, such as `images/**/*.jpg`.  In this case, the files that match the pattern are available to the Wasm code, at the same paths as they are in your file system. For example, if the glob pattern matches `images/photos/lake.jpg`, the Wasm module can access it using the path `/images/photos/lake.jpg`.  Glob patterns are relative to the directory containing `spin.toml`, and must be within that directory.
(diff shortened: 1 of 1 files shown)


content/spin/writing-apps.md
```diff
@@ -263,7 +263,7 @@ files = [ "images/**/*.jpg", { source = "styles/dist", destination = "/styles" }
 
 The `files` field is an array listing the files, patterns and directories you want to include. Each element of the array can be:
 
-* A glob pattern, such as `images/**/*.jpg`.  In this case, the files that match the pattern are available to the Wasm code, at the same paths as they are in your file system. For example, if the glob pattern matches `images/photos/lake.jpg`, the Wasm module can access it using the path `/images/photos/lake.jpg`.  Glob patterns are relative to the directory containing `spin.toml`, and must be within that directory.
+* A glob pattern, such as `images/**/*.jpg`.  In this case, the files that match the pattern are available to the Wasm code, at the same paths as they are in your file system. For example, if the glob pattern matches `images/photos/lake.jpg`, the Wasm module can access it using the path `images/photos/lake.jpg`.  Glob patterns are relative to the directory containing `spin.toml`, and must be within that directory.
```
itowlson, line 266: In the guest, globbed files retain their position relative to the app directory, but transplanted to be relative to `/`... which is also the guest working directory.  So the original text is right... and this text is also right (provided the guest doesn't change current directory, which I don't know if that is even a thing for Wasm).

Pointing people at the relative path is probably better because the whole point of globs was that you would get the same behaviour in native (e.g. unit tests) as in Wasm.  And the path in the original text would _not_ work in native.

Pardon my ramble.  Just trying to think this through but I think I agree!

karthik2804 (approved):
LGTM

itowlson (approved):
A mathematician was giving a lecture.  He wrote an equation on the blackboard, remarking to his students, "Now, it is obvious that ...".  Then he stopped, peered at the blackboard and asked himself out loud, "_Is_ it obvious?"  The lecturer then fell silent, clearly thinking furiously.  After some ten minutes or so, he emerged from his trance, said, "Yes. It _is_ obvious," and continued with the lecture.

Anyway that is basically my comment, except for the falling silent bit.

content/spin/writing-apps.md
kate-goldenring: This makes sense! Thanks Ivan!

PR #443, opened 2023-03-09
feat(check-broken-links.sh): add initial website health check

vdice:
- add initial website health check to fail fast if root url isn't returning 200
- also fix behavior where blc would exit non-zero but overall check passes

### DIFF
--- a/.build/check-broken-links.sh
+++ b/.build/check-broken-links.sh
@@ -9,6 +9,9 @@ require() {
 }
 
 cleanup() {
+  if [[ "$?" == "124" ]]; then
+    echo "timeout exceeded waiting for the developer portal to run"
+  fi
   echo "stopping process for $$"
   cleanup_with_child $$ false
 }
@@ -50,11 +53,20 @@ export PREVIEW_MODE=1
 npm run spin >/dev/null 2>&1 &
 
 ## wait for portal to be up and running
-timeout 60s bash -c 'until curl --silent -f http://localhost:3000 > /dev/null; do sleep 2; done'
+export RESP_CODE="$(mktemp)"
+timeout 10s bash -c 'until curl -o /dev/null -s -w "%{http_code}\n" http://localhost:3000 > $RESP_CODE; do sleep 2; done'
+
+echo "checking website health via http://localhost:3000"
+
+[[ "$(< $RESP_CODE)" == "200" ]] && \
+  (echo "Success: website returned a 200 response code") || \
+  (echo "Failure: unexpected response code: $(< $RESP_CODE)" && exit 1)
 
 echo "starting link checker"
 
-## Run the broken link checker          
+## Run the broken link checker
+report="$(mktemp)"
+blc_error=false
 blc --recursive http://127.0.0.1:3000                                                                                                                                                       \
                                                                                                                                                                                             \
     `## returns 403`                                                                                                                                                                        \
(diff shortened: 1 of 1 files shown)


.build/check-broken-links.sh
```diff
@@ -77,8 +88,10 @@ blc --recursive http://127.0.0.1:3000
     --exclude 'https://www.instagram.com/fermyontech/'                                                                                                                                      \
     --exclude 'https://www.linkedin.com/company/fermyon/'                                                                                                                                   \
     --exclude 'https://support.google.com/webmasters/answer/7552505'                                                                                                                        \
-    --exclude 'https://support.google.com/webmasters/answer/9008080?hl=en'| tee /dev/stderr | grep "├─BROKEN─" > broken_links || true
-   
+    --exclude 'https://support.google.com/webmasters/answer/9008080?hl=en' | tee "${report}" || \
```
vdice, line 91: There's probably other ways to fix this, but with the previous logic, false positives were possible.  For instance, if the `blc` process itself exited non-zero, we'd not catch it (per the `|| true`), leading to scenarios like this:
```
$ npm run test
...
Getting links from: http://127.0.0.1:3000/foo
Error: HTML could not be retrieved
All the links are valid!
stopping process for 32479
cleanup_with_child 32479
stopping child 32544 (parent: 32479)
cleanup_with_child 32544
stopping child 32633 (parent: 32544)
cleanup_with_child 32633
stopping child 32698 (parent: 32633)
cleanup_with_child 32698
stopping child 32773 (parent: 32698)
cleanup_with_child 32773
stopping child 32774 (parent: 32773)
cleanup_with_child 32774
stopping self 32774
stopping self 32773
stopping self 32698
stopping self 32633
stopping self 32544

$ echo $?
0

Comment (karthik2804) [reply .build/check-broken-links.sh:91]: good catch!

File: .build/check-broken-links.sh [hunk 2]
```diff
@@ -50,12 +53,20 @@ export PREVIEW_MODE=1
...
 ## wait for portal to be up and running
-timeout 60s bash -c 'until curl --silent -f http://localhost:3000 > /dev/null; do sleep 2; done'
+timeout 10s bash -c 'until curl --silent -f http://localhost:3000 > /dev/null; do sleep 2; done'
+
+echo "checking website health"
+
+resp_code=$(curl -o /dev/null -s -w "%{http_code}\n" 127.0.0.1:3000)
+[[ "${resp_code}" == "200" ]] && \
+  (echo "Success: website returned a 200 response code") || \
+  (echo "Failure: unexpected response code: ${resp_code}" && exit 1)
```
karthik2804, line 63: Just have a curious question. would we not be able to get the response code from the curl while we wait for the website to come up?

karthik2804 (approved):

.build/check-broken-links.sh
vdice: Oh, let me spend a little more time on this.  Now we don't get the nice broken link report when this is the cause for the blc failure. 😅

.build/check-broken-links.sh
vdice: Good point, with the `-f` flag above, it will just timeout if the site doesn't return a successful response code.  We could leave that as is and continue to bubble up generic "timeout exceeded waiting for the site to come up" _or_ we might consider removing the `-f` flag and proceeding with this check and fail with an error mentioning the specific non-200 error code being returned.  WDYT?

vdice:
Let me put this in Draft and revisit tomorrow -- I've had fun with this one but need to attend to some other things for today.  Thanks for collaborating, @karthik2804!

tpmccallum:
Thanks so much, @vdice and @karthik2804; looking forward to this coming out of draft; love this work!!!

.build/check-broken-links.sh
vdice: Ok, sorry for the delay.  Now ready for a re-review @karthik2804!

karthik2804 (approved):
LGTM!

PR #446, opened 2023-03-10
Link language guides to getting started docs

### LINKED ISSUE #438: Language-specific pages don't link back to install instructions
I would like to be able to direct JS devs directly to the JS page, and Python devs directly to the Python page.

But right now, if I do there is no link back to installing Spin. It starts by assuming the user has Spin installed. I think we could fix with brief links in each of these docs. And I feel like it'd be great to be able to send devs directly to the language-of-their-choice guide so they can see right away what they can do without leaving their comfort zone.

Exmaple:
https://developer.fermyon.com/spin/javascript-components

itowlson:
Fixes #438.

@technosophos I linked to the quick start rather than the install page, because the quick start includes not only install instructions, but also how to get the templates and tools (e.g. Rust target, JS plugin).  The actual install page is more about advanced installation options.

I have **not** attempted to update any of the page bodies - that is on the list but I want to decouple it from this PR.

The current note format does mean the language guides now start with a bit of a wall of grey:

image

Longer term it might be nice to figure out a way to moderate this or break it up a bit.

@karthik2804 Is it possible to link to a page with a query string that sets the active `sdk-type` tab?  If not, it might be a nice feature though it is **not required** for this and should not be taken as a high priority!


itowlson:
(I know the screenshot has a typo about having Python installed; it's fixed in the PR.)

tpmccallum (approved):

karthik2804:
@itowlson yes it is possible to use the query parameter to set the SDK language and it is quite easy to implement although it isn't a current part of it. I will add it to my list of low priority things to do.

PR #449, opened 2023-03-11
fix the blockquote light mode shading on docs

flynnduism:
Small bug fix following #428 - the blockquotes in light mode were too dark to be legible. This PR corrects the color scheme.

before/after preview of change:

light-blockquote

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -1813,8 +1813,9 @@ html.dark-theme .search-modal-container .result-section-container .suggested-pro
 }
 
 .documentation .content section blockquote p {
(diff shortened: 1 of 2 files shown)


karthik2804 (approved):
The changes look good. The `static/data.json` has been included, can we just revert the changes to it?

itowlson:
I like this - thanks!  However, there seems to be a small knock-on effect on the `details` shortcut:

image

For me, the top box (`Why do I need a Spin interface?`) is now _so_ light that it no longer stands out as a box at all (whereas the second one, below the tables, is perfect).

PR #456, opened 2023-03-14
Move build toolchain information to where it is needed

tpmccallum:

Screenshot 2023-03-14 at 1 54 59 pm

Screenshot 2023-03-14 at 2 02 26 pm

Screenshot 2023-03-14 at 1 56 20 pm

### DIFF
--- a/content/spin/install.md
+++ b/content/spin/install.md
@@ -33,17 +33,6 @@ It's highly recommended to add Spin to a folder, which is on your path, e.g.:
 $ sudo mv spin /usr/local/bin/
 ```
 
-> Please note: On a fresh Linux installation, you will also need the standard build toolchain
-(`gcc`, `make`, etc.), the SSL library headers, and on some distributions you may need `pkg-config`.
-
-On Debian-like distributions, including Ubuntu, you can install the standard build toolchain with a command like this:
-
-<!-- @selectiveCpy -->
-
-```bash
-$ sudo apt-get install build-essential libssl-dev pkg-config
-```
-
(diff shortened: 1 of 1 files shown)


content/spin/install.md
```diff
@@ -108,6 +97,14 @@ $ cd spin && make build
 $ ./target/release/spin --help
 </code></pre>
 
+> Please note: On a fresh Linux installation, you will also need the standard build toolchain (`gcc`, `make`, etc.), the SSL library headers, and on some distributions you may need `pkg-config`. For example, on Debian-like distributions, including Ubuntu, you can install the standard build toolchain with a command like this:
```
itowlson, line 100: This would also be needed for the Using Cargo to Install Spin method.

itowlson, line 100: I would suggest rewording away from "a command like this" to "this command."

itowlson (changes requested):
Just need to copy it to one more place I think

content/spin/install.md
tpmccallum: Done

tpmccallum: I might have to re-word a little to cater for Windows users (as not to make it too Linux focussed when referring to cargo)

itowlson: Ooh good catch

itowlson (approved):

content/spin/install.md
tpmccallum: As discussed, I agree, catering for the alternate platforms can go over to the contributing to Spin page, which is capable of diving into a bit more detail and providing good context. about installing from source.

PR #457, opened 2023-03-14
Please review: Language Support Overview

tpmccallum:
This is ready for review.

Please comment at [the file level itself](https://github.com/fermyon/developer/pull/457/files#r1138179312) any feedback is greatly appreciated. If this approach is sound then I can go ahead and populate the other language tabs.

- [x] Ensure all links to content on the same domain are relative and not absolute; to save future issues/broken links.
- [x] **Rename the actual `.md` file** to the name that has been decided on for the page title i.e. "Language Support Overview" means the page must be named `language-support-overview.md`. Changing the file name prematurely upsets the conversation/review/comments which are currently very valuable so let's change the file name just before we push.

Screenshot 2023-03-17 at 8 57 10 pm

Screenshot 2023-03-17 at 8 57 45 pm

- [x] Needs to be added to the Spin menu in the `.hbs` file (need to choose which sub heading on the Spin menu this is best suited to)


karthik2804:
We could potentially use tables inside each language?

mikkelhegn:
I'd recommend we create a custom template for this. @flynnduism - can you help with the layout of this guide table?

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,533 @@
+title = "Language Guide Overview"
```
itowlson, line 1: This seems like more of a feature matrix than a "language guide overview."  Maybe "Language Support for Spin" or "Language Status"?

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,533 @@
...
+enable_shortcodes = true
+[extra]
+
+---
+
+The following code block outlines, in the context of the Spin framework, what languages have SDK support, and what SDK features are available in each of those languages:
+
+{{ tabs "sdk-type" }}
+
+{{ startTab "C#/.NET"}}
```
itowlson, line 13: Our other `sdk-types` tab sets have the order and terminology: Rust | TypeScript | Python | TinyGo

We should keep the names consistent (I don't care about JavaScript vs TypeScript, or Go vs TinyGo, as long as it's consistent), so that a user's choice of language travels with them.

I would also suggest keeping the order consistent both for looking organised, and for muscle memory.  Languages that aren't yet everywhere, such as C#, would go at the end.

Aside: at present, I would _never_ put C# first, _anywhere._   The SDK is not yet mature enough to be placed in the most prominent position.

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,533 @@
+title = "Language Guide Overview"
+template = "common_main"
+date = "2022-01-01T00:00:01Z"
+enable_shortcodes = true
+[extra]
+
+---
+
+The following code block outlines, in the context of the Spin framework, what languages have SDK support, and what SDK features are available in each of those languages:
```
itowlson, line 9: Suggest something like "This page" rather than "The following code block," and maybe rewording away from "in the context of the Spin framework".  How about something like "This page contains information about language support for Spin features."

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,533 @@
...
+
+Blog Article[s]:
+
+**Feedback**
+
+{{ blockEnd }}
+
+{{ startTab "Rust"}}
+
+**SDK: Outbound HTTP**
```
itowlson, line 433: Suggest dropping the `SDK` prefix.  (It's all SDK.)

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,533 @@
...
+
+Local: Yes
+
+Cloud: Yes
```
itowlson, line 437: Local vs cloud is not a language issue, it's a feature issue.  There's no circumstance in which a SDK would support a feature locally but not in cloud, or vice versa.  The SDK simply supports the feature or doesn't.

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,533 @@
...
+
+Documentation: [https://developer.fermyon.com/spin/rust-components.md#sending-outbound-http-requests](https://developer.fermyon.com/spin/rust-components.md#sending-outbound-http-requests)
```
itowlson, line 439: It might be safer to use relative links for within the documentation?  That would allow us to detect at CI time if a destination page has moved.

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,533 @@
...
+
+Code Examples: [https://github.com/fermyon/spin/tree/main/examples/rust-key-value](https://github.com/fermyon/spin/tree/main/examples/rust-key-value)
+
+Blog Article[s]:
+
+**SDK: Outbound MySQL**
+
+Local: Yes
+
+Cloud: Yes (non-hosted)
```
itowlson, line 461: I'm not sure what this is saying.  MySQL is always non-hosted, whether local or cloud; phrasing it this way suggests that local _is_ hosted.

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,533 @@
...
+
+Documentation: [https://developer.fermyon.com/spin/rust-components.md#storing-data-in-redis-from-rust-components](https://developer.fermyon.com/spin/rust-components.md#storing-data-in-redis-from-rust-components)
+
+Code Examples: [https://github.com/fermyon/spin/tree/main/examples/rust-outbound-redis](https://github.com/fermyon/spin/tree/main/examples/rust-outbound-redis)]
+
+Blog Article[s]:
+
+**Trigger: Custom (Cron)**
```
itowlson, line 493: Either "custom triggers" (for general behaviour) or "the cron trigger" (for custom triggers with specific support).

I would suggest not talking about cron in this context - we have, I think, a timer sample, but not a working cron trigger that people can use.

content/common/language-guide-overview.md
itowlson, line 437: +1 to the table suggestion for this.  What about something like:

| Feature | SDK Supported? | Learn More |
|-----|-----|-------|
| Triggers |
| HTTP trigger | Supported | [Reference](/spin/foo)<br>[Tutorial](/spin/foo)<br/>[Routing blog](https://example.com/r)<br />[Basic Example](https://example.com/blah) |
| Redis trigger | Supported | [Reference](/spin/foo)<br>[Tutorial](/spin/foo)<br/>[Dead Letter Queue Example](https://example.com/blah) |
| APIs |
| Outbound HTTP | Supported | [Reference](/spin/foo)<br>[Tutorial](/spin/foo)<br/>[Dog Facts Demo](https://example.com/blah) |
| MySQL | Unsupported | - |

(of course the Learn More could be split into multiple columns if we want to be precise about docs vs blogs vs examples)

(also we might need to crack multi-column cells to do the triggers vs APIs heading but that's not critical)

itowlson: I am also wondering if we accompany this with an "API Guide" which has the same information but tabbed by API rather than by language, and each tab body talks about cloud and language support and examples, and cross links between that page and this.  But maybe the API pages already cover that?

Which also makes me wonder: should this info be on each language page, either instead of or as well as in the tabbed block?  We don't a reader to have to jump back here when they are already on the language page...?

itowlson (commented):
Apologies if I got too bogged down in wording for this early stage, but there are some more substantive comments in there too, I promise...!

content/common/language-guide-overview.md
tpmccallum: This is great feedback @itowlson, thank you!

**Order**
I initially just wrote the page out in alphabetical order, on purpose, so there was no bias towards any language. I am hearing you, and it does make sense to put more mature offerings at eye level; for sure. So essentially duplicating the following visual that already exists elsewhere (and then adding any new languages at the end (or on the right-hand side).

Screenshot 2023-03-15 at 8 29 55 am

**Consistency**
Thanks for the `sdk-types` consistency comment also; I 100% agree with this, just wasn't sure which way to go but I guess the existing `sdk-types` in existing code blocks are already setting the standard so it makes sense to match these with what already exists. If we change our minds (for whatever reason; then we reflect those changes throughout the whole developer docs for consistency).

content/common/language-guide-overview.md
tpmccallum: Thanks @itowlson,

I like "Language Support Overview" if everyone is ok with that. Happy to update if/when the discussion continues.

content/common/language-guide-overview.md
tpmccallum: Sounds great @itowlson 
Updated in next push to this branch.
I like how the text matches the page's title (which will also eventually match the page's name); succinct and to the point!

content/common/language-guide-overview.md
tpmccallum: I have dropped the `SDK:` prefixes in each spot. Shall we leave the `Trigger:` prefix on the other ones? Which is currently the case.

content/common/language-guide-overview.md
tpmccallum: I have added an item to the checklist for this PR to ensure that relative URLs are used where appropriate. Thanks, good suggestion.

content/common/language-guide-overview.md
itowlson: I think it would read better as a "Triggers" subhead (but then we'd need a corresponding subhead for the APIs), or as "HTTP trigger".

content/common/language-guide-overview.md
tpmccallum, line 1: Update to the Rust tab only for feedback purposes.
Is the Rust tab on the right path?

Screenshot 2023-03-16 at 4 21 48 pm

rajatjindal: should we add a ✅ instead of text `supported`?

mikkelhegn: This looks great Tim!

macolso: Looks great! Small knit - would it make sense to have that second header say SDK APIs , or is that redundant given its a language matrix?

itowlson: @macolso It's all SDK though - triggers too.

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,448 @@
...
+| Redis | Supported | [Documentation](https://developer.fermyon.com/spin/redis-trigger)<br/> [Code example](https://github.com/fermyon/spin/tree/main/examples/redis-rust)<br/>Blog|
+| Custom | Supported | [Documentation](https://developer.fermyon.com/spin/extending-and-embedding)<br/>Code example<br/>Blog|
+| **APIs** |
+| Outbound HTTP | Supported | [Documentation](https://developer.fermyon.com/spin/rust-components.md#sending-outbound-http-requests)<br/>[Code example](https://github.com/fermyon/spin/tree/main/examples/http-rust-outbound-http)<br/>Blog|
+| Key Value Storage | Supported | [Documentation](https://developer.fermyon.com/spin/kv-store.md)<br/>[Code example](https://github.com/fermyon/spin/tree/main/examples/rust-key-value)<br/>Blog|
+| Outbound MySQL | Supported | [Documentation](https://developer.fermyon.com/spin/rdbms-storage#using-mysql-and-postgresql-from-applications)<br/>[Code example](https://github.com/fermyon/spin/tree/main/examples/rust-outbound-mysql)<br/>[Blog](https://www.fermyon.com/blog/building-a-social-app-with-spin-1) |
```
itowlson, line 24: I know this is not final wording or anything but I would drop "outbound" from MySQL and PostgreSQL.

content/common/language-guide-overview.md
```diff
@@ -0,0 +1,448 @@
...
+{{ tabs "sdk-type" }}
+
+{{ startTab "Rust"}}
+
+| Feature | SDK Supported? | Learn More |
+|-----|-----|-------|
+| **Triggers** |
+| HTTP | Supported | [Documentation](https://developer.fermyon.com/spin/http-trigger)<br/> [Code example](https://github.com/fermyon/spin/tree/main/examples/http-rust-outbound-http)<br/>Blog|
+| Redis | Supported | [Documentation](https://developer.fermyon.com/spin/redis-trigger)<br/> [Code example](https://github.com/fermyon/spin/tree/main/examples/redis-rust)<br/>Blog|
+| Custom | Supported | [Documentation](https://developer.fermyon.com/spin/extending-and-embedding)<br/>Code example<br/>Blog|
```
itowlson, line 20: Again I know this is not a review for content but custom triggers are not "supported" in Rust or anywhere else - almost by definition because once we support them they are no longer custom!  _grin_

(Or, per the link, is this more about _authoring_ custom triggers?  If so, maybe we should move the line to an "extensibility" or "other features" section - keep this section for using triggers _in applications._)

itowlson (commented):
I really like the look of this!  The comments are wording ones for when you fill it out, not on the overall structure.

content/common/language-guide-overview.md
itowlson: @rajatjindal If we do that I'd go for "as well as" instead of "instead of."  Thinking partly of screen readers, but also for the inevitable day when we need to say "supported except for..."!

tpmccallum: [Tim M.] Had a great chat with @mikkelhegn this morning and we made a decision that it would be great to condense the table (both from a height and width perspective). This update involves making the table in each tab 2 columns. The first column that has the name of the SDK/Trigger etc. is hyperlinked to the documentation of that SDK/Trigger etc. and will open on the correct tab (at the destination page; thanks to session/local-browser storage). The second column indicates if the SDK/Trigger etc. is supported.
This has worked well as a test on the Rust tab; as we can see it renders really well on the latest iPhone, for example.

Screenshot 2023-03-17 at 1 19 33 pm

The idea moving forward is to have an official piece of documentation available (on the official developer.fermyon.com site) which will itself provide code examples (or links to code examples) and links to other material such as blog posts on the topic and so forth. This keeps the Language Support Overview very clean and condensed.

content/common/language-guide-overview.md
tpmccallum: Ah, I see, I will add another section to separate the different feature type.

content/common/language-guide-overview.md
itowlson: Nice!

tpmccallum: After filling out all of the tabs there was one apparent "Supported" but not documented feature. Is it possible to take code from [the recent Python blog](https://www.fermyon.com/blog/spin-python-sdk) that mentions key/value in both `.toml` and `app.py` and then add a Python tab to the [spin key/value documentation page](https://github.com/fermyon/developer/blob/main/content/spin/kv-store.md). Or will that cutting and pasting be misleading, give that more work has to take place in order to create the Python key/value storage application in the first place (compared to what we have documented on Rust i.e. just using a single template command to kick things off)?

tpmccallum: I have added a new PR which adds a Python tab to the code blocks in the key value documentation (also added a reference to the official blog post that showed how to do key value storage). This now means we don't have a supported feature without a link in the left column (undocumented).
https://github.com/fermyon/developer/pull/483

tpmccallum:
The latest commit is ready for review.

mikkelhegn (approved):

PR #463, opened 2023-03-14
make comments brighter

### LINKED ISSUE #420: Comments in code blocks should be prominent, or at least visible
I wrote a comment in this code block, to draw the reader's attention to the bit I wanted to highlight.

image

But the comment is rendered as dark grey on black, making it very hard to see.

Comments aren't something that should be discreetly receded; _particularly_ in examples, they are there to point the reader at the most important bits.

karthik2804:
Changes the color of the comments to `#bbb` to make it more visible. We could also bump the font-weight to bring more attention to it.

fixes #420 

image

### DIFF
--- a/static/sass/styles.scss
+++ b/static/sass/styles.scss
@@ -173,6 +173,10 @@ TABLE OF CONTENTS
 			code {
 				padding: 1.125em 1.25rem !important;
 				margin-right: 2.75rem;
+
+				.hljs-comment {
+					color: #bbb !important;
+				  }
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


static/sass/styles.scss
```diff
@@ -173,6 +173,10 @@ TABLE OF CONTENTS
 			code {
 				padding: 1.125em 1.25rem !important;
 				margin-right: 2.75rem;
+
+				.hljs-comment {
+					color: #bbb !important;
+				  }
```
itowlson, line 179: Is this alignment right?!?!?!?!

itowlson (approved):
🎉 :shipit: 🎉 :shipit: 🎉 :shipit: 🎉

static/sass/styles.scss
karthik2804: yes that is nested `scss` indentation!

PR #464, opened 2023-03-14
auto detect os if there is no preference

### LINKED ISSUE #401: OS specific instructions should detect current OS
When reading documentation that is specific depending on the operating system used, the page should try to determine the OS and show the appropriate tab by default.

Currently, the first tab is always shown regardless of the OS:

image

karthik2804:
Auto detects the OS if the user has not set a preference. 

fixes #401


static/js/src/modules/multiTab.js
```diff
@@ -68,48 +70,65 @@ class multiTabContentHandler {
...
         }
-        if(updateLocalStorage) localStorage.setItem("toggleTabSelections", JSON.stringify(this.selectedTab))
+        if (updateLocalStorage) localStorage.setItem("toggleTabSelections", JSON.stringify(this.selectedTab))
+    }
+}
+
+function detectOS() {
+    let OS = null
+    let userAgent = navigator.userAgent.toLowerCase()
+    switch (true) {
```
itowlson, line 120: I haven't seen this idiom and it looks really strange to me but if it works it works - TIL!

itowlson (approved):

static/js/src/modules/multiTab.js
karthik2804: I am a bit lost as to what you mean here.

itowlson: No action, I am just expressing that I learned something new from this PR.

karthik2804: ah!

karthik2804: Could have written it with if statements too!

PR #468, opened 2023-03-15
add cli stability table

michelleN:


content/common/cli-reference.md
```diff
@@ -1208,3 +1209,28 @@ TRIGGER OPTIONS:
...
+| `spin add`                      | Stable       |
+| `spin build`                    | Stable       |
+| `spin new`                      | Stable       |
+| `spin plugins`                  | Stable       |
+| `spin templates`                | Stable       |
+| `spin up`                       | Stable       |
+| `spin cloud deploy`             | Stabilizing  |
+| `spin cloud login`              | Stabilizing  |
+| `spin registry`                 | Stabilizing  |
+| `spin bindle <all subcommands>` | Experimental |
```
itowlson, line 1232: Deprecated rather than experimental I'd suggest

content/common/cli-reference.md
```diff
@@ -1208,3 +1209,28 @@ TRIGGER OPTIONS:
...
 
 {{ blockEnd }}
+
+### CLI Stability Table
+
+CLI commands have three phases that indicate levels of stability:
+
+- `experimental`: These commands are experiments and may or may not be available in later versions of the CLI.
+- `stabilizing`: These commands have moved out of the `experimental` phase and we are now in the active process of stabilizing them. This includes updating flags, command output, errors, and more.
+- `stable`: These commands have moved out of the `stablizing` phase and will not change until the next major version release.
```
itowlson, line 1219: "Will not change" might be too strong - they won't change _in backward-incompatible ways._  For example, a minor release could add new optional flags (provided the behaviour if omitted didn't change).

content/common/cli-reference.md
itowlson, line 1232: `<all subcommands>` isn't stated on `spin plugins` and `spin templates` - what is our preferred convention?

content/common/cli-reference.md
```diff
@@ -1208,3 +1209,28 @@ TRIGGER OPTIONS:
...
+
+| Command                         | Stability    |
```
itowlson, line 1221: Is it worth saying which version of Spin the table refers to?

content/common/cli-reference.md
```diff
@@ -1208,3 +1209,28 @@ TRIGGER OPTIONS:
...
+| ------------------------------- | ------------ |
+| `spin add`                      | Stable       |
+| `spin build`                    | Stable       |
+| `spin new`                      | Stable       |
+| `spin plugins`                  | Stable       |
+| `spin templates`                | Stable       |
+| `spin up`                       | Stable       |
```
itowlson, line 1228: What about the `--bindle` flag to `spin up` - do we intend to remove that before 1.0, retain it for the duration of 1.x, or exclude it from the 'stable' state?  It is currently hidden and prints a deprecation warning.

Related: using parcels as component module sources in the manifest.

content/common/cli-reference.md
michelleN: agreed

michelleN: I think we should list all commands since that list might change over time. Better to be explicit.

`spin bindle <prepare | push>`
`spin plugins <list | uninstall...etc>`

itowlson: That's certainly the safest approach.  There's a risk it could bulk out the table (commands like deploy and up getting lost in all the plugin and template ones) but we should definitely try it that way, we can always iterate if we don't like it!

michelleN: updated.

content/common/cli-reference.md
michelleN: updated.

content/common/cli-reference.md
michelleN: updated.

content/common/cli-reference.md
michelleN: Added the spin version tab like the other sections in this doc.

michelleN:
> One question/suggestion - Do we want to move the table to the top of the page? I feel like it is hidden way far too down now.

@karthik2804 I added a link to the stability table in the first paragraph to make it more accessible. Thanks for the suggestion.

content/common/cli-reference.md
michelleN: Good point. I think we want to ensure `up` is stable. Continuing this conversation in slack.

PR #471, opened 2023-03-15
up should be stabilizing until we figure out what path we want to go with bindle options

michelleN:
Merged the CLI stability table but realized we should have the status of `spin up` be `Stabilizing` until we reconcile @itowlson's question about the bindle options.


tpmccallum (approved):
Thanks @michelleN 
Happy to err on the side of caution.

tpmccallum:
Sorry this clashes with the heavy changes made this morning Re: table formatting `|`
Will ask this in the Spin channel and then see if we need to update this or not.

PR #477, opened 2023-03-16
Updating Windows install instructions

mikkelhegn:

Small clarification about WSL and using Windows natively.

### DIFF
--- a/content/spin/install.md
+++ b/content/spin/install.md
@@ -13,7 +13,7 @@ keywords = "install"
 
 ## Installing Spin
 
-Spin runs on Linux (amd64 and arm64), macOS (Intel and Apple Silicon), and Windows with WSL2 (amd64).
+Spin runs on Linux (amd64 and arm64), macOS (Intel and Apple Silicon), and Windows (amd64).
 
 {{ tabs "os" }}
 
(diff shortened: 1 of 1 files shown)


karthik2804:
@itowlson is there a requirement for WSL2 over WSL1 or does it work with both?

content/spin/install.md
```diff
@@ -79,7 +79,9 @@ To install the canary version of spin, you should pass the argument `-v canary`.
 
 {{ startTab "Windows"}}
 
-On Windows systems, you can download <a href="https://github.com/fermyon/spin/releases/latest" class="spin-install" id="spin-install-windows">the Windows binary release of Spin</a>.
+If you want to use WSL (Windows Subsystem for Linux), please follow the instructions for using Linux.
```
itowlson, line 82: This might sit more happily in a note/blockquote so that it is set aside from the actual-Windows instructions.  I'd also move it below the actual-Windows instructions (and revert that paragraph).  The reason is that WSL users generally know they are using WSL and will tend to gravitate to the Linux instructions, so this tab should focus on the actual-Windows case as its primary concern, rather than as the "else" part of an if statement _grin_

itowlson (changes requested):

karthik2804:
We might want to make it more specific by mentioning it at WSL2?

content/spin/install.md
mikkelhegn: Like this?
image

itowlson: I was thinking inside the Windows tab, but that is probably clearer - thanks!

(I would still drop the reference to "natively.")

mikkelhegn: Going with this, the block thing made it stand out far too much:
image

tpmccallum (approved):

tpmccallum:
Just performing some checks & LGTM
Screenshot 2023-03-17 at 9 25 30 pm
Screenshot 2023-03-17 at 9 24 55 pm

itowlson (approved):
LGTM - thanks!

PR #480, opened 2023-03-16
allow toggling multitabs from query params

karthik2804:
This PR allows specifying what tabs to be selected on page load using query parameters. The query parameters for the tab names are prefixed with `multitab`. An example URL would be
```
http://localhost:3000/spin/quickstart?multitab_os=Linux&multitab_sdk-type=TinyGo
```


itowlson (approved):
LGTM!

PR #482, opened 2023-03-17
Update broken links on quickstart page

tpmccallum:

### DIFF
--- a/content/spin/quickstart.md
+++ b/content/spin/quickstart.md
@@ -802,6 +802,6 @@ Congratulations! You just created, built and ran your first Spin application!
 
 ## Next Steps
 
-- Learn more about [writing Spin components and manifests](writing-apps)
-- Learn how to [build your Spin application code](build)
+- Learn more about [writing Spin components and manifests](/spin/writing-apps)
+- Learn how to [build your Spin application code](/spin/build)
 - Try deploying a Spin application to the [Fermyon Cloud](/cloud/quickstart)


itowlson (approved):
I would like to leave the issue open to prompt us to investigate.  We don't know why these work locally, or if any other links on the site are broken; we need to understand as well as emergency patching.  (But yes emergency patch for now!)

karthik2804:
I have it figured out. If there is a trailing `/`, the link is broken.

itowlson (changes requested):
From what @karthik2804 has found we need to solve the problem elsewhere

itowlson:
@tpmccallum Sorry for leaving a request changes - there is no point papering over these two specific links when the issue is global and could hit any page that has the slashy link, so I wanted to put this on hold while we figure it out.

karthik2804:
The only places with trailing slashes are in the search box.  We should possibly not be using those relative URLs given any external website can link that way as well. 

https://github.com/fermyon/developer/blob/1754705721cca42749482642411b40ccf4bedecf/static/js/src/modules/search.js#L179


image

itowlson:
Can we fix this in Bartholomew @karthik2804 or do we need to work around it by authoring absolute links from now on?

karthik2804:
The only way it can be fixed in Bartholomew is to translate the relative URLs to absolute URLs. I would have to look into that a little bit before I can confirm. But yes that should be possible.

tpmccallum:
In reference to the above comment, https://github.com/fermyon/developer/pull/482#issuecomment-1473110820, this will be close in lieu of another different approach to fix.

PR #496, opened 2023-03-21
docs(http-trigger.md): link rendering fixes and minor updates

vdice:
- Fixes issue where some links weren't clickable (perhaps because there was a trailing `.` or `:`?)  Please advise if there is a preferred approach to having them render correctly.
- Add'l minor fixes/updates

Before:
hyperlinks-broke

After:
hyperlinks-fixed


karthik2804 (approved):

karthik2804 (approved):

itowlson:
@karthik2804 is there a way to fix this in Bartholomew?  Trailing punctuation doesn't normally prevent auto-linking in other Markdown environments I've used...?

karthik2804:
I am not sure that I follow. Do you mean syntax like `[example](www.example.com).` They do work as expected.

vdice:
@karthik2804 I think @itowlson meant the way the links were before this fix.  They were raw, valid links eg `https://github.com/fermyon/spin/blob/main/wit/ephemeral.` but my theory was the hyperlink functionality broke when punctuation was involved, such as the `.` at the end there.  Bartholomew issue or maybe js logic issue for this website?

karthik2804:
@vdice @itowlson, it makes sense now. It is due markdown parser we use. it does not auto-detect URLs even in cases of lack of punctuation. I can look into it further. But from a quick glance, it looks like the syntax for the raw URL is `<URL_HERE>` that is the URL is enclosed in angled brackets.

PR #501, opened 2023-03-23
Remove `--follow-all` from docs

tpmccallum:

Tested this out; `spin up` alone will print the data published from the Redis CLI (as per this example) without the need for `--follow-all`.

This is the only occurrence left; checked using `find content -type f -exec grep -H '\--follow\-all' {} \;`


tpmccallum:
Screenshot 2023-03-23 at 12 19 33 pm
Screenshot 2023-03-23 at 12 19 24 pm

michelleN (approved):

PR #503, opened 2023-03-23
Adding Redis trigger authentication methods

tpmccallum:

### DIFF
--- a/content/spin/redis-trigger.md
+++ b/content/spin/redis-trigger.md
@@ -7,6 +7,7 @@ url = "https://github.com/fermyon/developer/blob/main//content/spin/redis-trigge
 
 ---
 - [Specifying an Application as Redis](#specifying-an-application-as-redis)
+- [Authenticating with Redis](#authenticating-with-redis)
 - [Mapping a Channel to a Component](#mapping-a-channel-to-a-component)
 - [Redis Components](#redis-components)
 	- [The Message Handler](#the-message-handler)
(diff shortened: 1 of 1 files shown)


content/spin/redis-trigger.md
```diff
@@ -34,11 +35,17 @@ The Redis trigger also requires an `address` field.  This is the address of the
 
 > If you create an application from a Redis template, the trigger will be already set up for you.
 
-By default, Spin does not authenticate to Redis. You can work around this by providing a password in the `redis://` URL.  For example: `address = "redis://:p4ssw0rd@localhost:6379"`
+## Authenticating with Redis
 
-> Do not use passwords in code committed to version control systems.
+By default, Spin does not authenticate to Redis. However, you can authenticate, using Spin's:
+- [config at the application level](https://developer.fermyon.com/spin/dynamic-configuration.md#custom-config-variables),
```
vdice, line 41: I think we'll need to revise this list or reword, as these aren't mutually exclusive but interoperate.  Here's my understanding of how it works today (perhaps tag @itowlson or other to check me here):

The first entry here, config variables, represent necessary config _declaration_ that must be added for providing config to a component.  The second entry is component config _declaration_ which may only reference config variable declarations (first entry).  Entries 3-5 represent config _provider_ approaches for providing values corresponding to config variables used by components.  The env var provider and the Vault provider are two different approaches to accomplish this.  Entry 5, the runtime config file, holds ... wait for it... config for the config provider so that it may provide config 😂.  As of today, I think the only `[[config_provider]]` that the runtime config file supports is the Vault config provider, until other options are added.

tpmccallum: Thanks for your comments @vdice 
I thought about this over the weekend and have a suggestion.
In light of this great feedback, perhaps we make this PR about upgrading the docs from only mentioning a static work-around i.e. `redis://password_etc`, to move into introducing the fact that Spin provides dynamic configuration.

What if I remove the itemised list and then mention (and link to) the dynamic configuration page (where the user can read the dynamic configuration content; rather than try to duplicate information here also)?

For example, 

```md
By default, Spin does not authenticate to Redis. However, the appropriate configuration can be achieved using Spin's [Dynamic and Runtime Application Configuration](https://developer.fermyon.com/spin/dynamic-configuration.md) options.
```

tpmccallum: Then we can elaborate on the dynamic configuration page regarding future iterations/updates.

tpmccallum: I have updated the file with the new content mentioned above; please let me know if you approve. 
Thanks
Tim

vdice: @tpmccallum I like the simplification.  However, I may have been mistaken here.  Interpolation of variables inside the redis `address` doesn't yet appear to be supported.  I've added a note to the applicable issue [here](https://github.com/fermyon/spin/issues/192#issuecomment-1485542933).  We may want to hold off on this PR for a bit to see if we get confirmation on behavior; if indeed dynamic/runtime config _can't_ yet be used for the redis address, we'll have to leave docs as is until support has been added.

tpmccallum: Thanks @vdice will park this PR for a while (until we know what is supported for sure).

vdice: @tpmccallum Indeed, injection of config to the `address` (or other fields) on the `trigger` section in the spin.toml manifest file aren't supported today.  Not sure when it will be prioritized at this point (for redis specifically, this is currently tracked by https://github.com/fermyon/spin/issues/192).  At this stage, I'd suggest we close this docs update.  Sorry for the misinformation here!

tpmccallum:
Closing this as it is outdated.

PR #517, opened 2023-03-28
Delete architecture.md

tpmccallum:


tpmccallum:
A dangling page which has outdated content and is not linked to by anything.
It is indexed in Google so this deletion is to remove it from being rendered in the sitemap.
If we need an architecture documentation page please let me know and we can write from scratch using the updated architecture and CLI etc.

tpmccallum:
Reference https://github.com/fermyon/developer/pull/517

rajatjindal (approved):

PR #528, opened 2023-03-30
do not run preview workflow on forks

rajatjindal:
do not run preview workflow on forks

### DIFF
--- a/.github/workflows/deploy-preview.yml
+++ b/.github/workflows/deploy-preview.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   deploy-preview:
     runs-on: ubuntu-latest
-    if: ${{ github.repository_owner == 'fermyon' }}
+    if: ${{ !github.event.pull_request.head.repo.fork }}
     name: Build and deploy
     steps:
       - uses: actions/checkout@v3


.github/workflows/deploy-preview.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   deploy-preview:
     runs-on: ubuntu-latest
-    if: ${{ github.repository_owner == 'fermyon' }}
```
radu-matei, line 10: Could you expand on why the existing check would not be enough?

Specifically if this is something we'd want to document to users of the action.

itowlson: @vdice I think the goal is more more "do not run preview workflow on PRs originating from forks".  From what I can tell, the issue that we're seeing is that preview seems to fail on _PRs created from forks_ (cf. https://github.com/fermyon/developer/actions/workflows/deploy-preview.yml - Tim's PRs from this repo succeed, mine and Kate's and Rajat's from our forks fail).  These PRs are in an owned-by-Fermyon repo and so pass this check; but I guess they don't have access to the secret, so the deployment fails.

vdice: Ah ha, thank you @itowlson.  TIL.  Indeed, I've just seen the same behavior on my PR for a different item, from my fork.

vdice (approved):
Glad to expedite this patch for a near-term fix and then we can huddle per @itowlson's [comment](https://github.com/fermyon/developer/pull/528#issuecomment-1490676057) re: next steps/broader strategy.

For this PR (or an immediate follow-up), do we want/need any mechanism to notify to PRs originating from forks to not expect a preview?  If this looks tricky (i.e. necessitates changes to the action itself), don't need to block this.

rajatjindal:
> Glad to expedite this patch for a near-term fix and then we can huddle per @itowlson's [comment](https://github.com/fermyon/developer/pull/528#issuecomment-1490676057) re: next steps/broader strategy.
> 
> For this PR (or an immediate follow-up), do we want/need any mechanism to notify PRs originating from forks to not expect a preview? If this looks tricky (i.e. necessitates changes to the action itself), don't need to block this.

this would need a token as well (to update comment on PR). We can just add a log msg in the Github action, but I don't think we can do anything more.

(technically we can have a webhook to update the comment on PR, but that won't work if it is a private repo). 

I'll give it more thoughts tomorrow to see if there is a workaround we can do.

PR #538, opened 2023-04-07
ref(kv): update kv doc

michelleN:
I found the timeline related references and structure of this document a little hard to understand when I was trying to just figure out how to use key/value storage. I re-worded the doc to reflect the current state so it is easier to both understand and maintain and I removed time related references like "in previous articles", etc.

Would appreciate feedback here.

### DIFF
--- a/content/spin/kv-store.md
+++ b/content/spin/kv-store.md
@@ -1,4 +1,4 @@
-title = "Persistent Data: Spin"
+title = "Persistent Data with Spin"
 template = "spin_main"
 date = "2023-02-21T00:00:00Z"
 enable_shortcodes = true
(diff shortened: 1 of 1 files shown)


itowlson:
Totally agree with making this document "in the present" but I think @macolso already has a rework in progress (#430) which also tackles the tutorial-vs-conceptual ambiguity and gets rid of the allusions to SQLite.  (There is also a Notion page for the storage docs rework but I can't find it again, sorry.)

itowlson:
@michelleN I think this is now covered by the current K/V documentation - is it okay to close it, or is there content in here that still needs to be brought in?  Thanks!

PR #573, opened 2023-04-26
Add centralized list of class names

tpmccallum:

This is a great suggestion from @karthik2804 to track the use of class names in the docs.

### DIFF
--- a/content/common/contributing-docs.md
+++ b/content/common/contributing-docs.md
@@ -152,7 +152,7 @@ The first step to implementing multi-tab code blocks is placing the `enable_shor
 The markup to create tabs in markdown is as follows 
 
 ```
-{{ tabs "Os" }}
+{{ tabs "os" }}
 
 {{ startTab "Windows"}}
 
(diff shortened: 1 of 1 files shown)


content/common/contributing-docs.md
```diff
@@ -197,6 +197,34 @@ test.sh
...
 
+**Note**: Existing documentation will already be using class names for code block `tabs` and `startTab` i.e. `{{ tabs "os" }}` and `{{ startTab "Windows"}}` respectively. Please consult the following `tabs` and `startTab` class names that are already in use (before creating your own). If you need to create a new class name (because one does not already exist) please add it to the list below as part of the pull request that contains your code block contribution.
+
+**tabs**:
+- `gh-interact`
+- `os`
+- `sdk-type`
+- `spin-version`
+
+**startTab**
```
itowlson, line 208: Is it worth mentioning under which `tabs` block these are typically used?

tpmccallum: Thinking hard about this one ...
I can't come up with a reason to associate the relationship between `tabs` and `startTab`. The main thing that leans toward not listing associations is that there may be cases where generic `startTab` names might fall under more than one `tabs` parent. Versions spring to mind initially. I tried using a table but then thought this list format was easier to scan (alphanumeric ordering) and also easier for a contributor to edit.

tpmccallum: TL;DR I am happy with just listing these separately.

itowlson: An example of why might be I am creating an `os` block, now what tabs should I have - Linux and Windows, okay, but macOS, Mac OS, Mac OS X, OS X, or Darwin?  For SDK type is it Go or TinyGo, C Sharp or C#?  Some kind of indicative "likely tabs" would help authors see this more readily.

Though realistically I always just copy tab blocks from somewhere else anyway _scurries away_

tpmccallum: Same :)

karthik2804 (approved):
LGTM - I just have a slight preference of listing common tab names but this is definitely good enough.

Issue #578, opened 2023-04-26
Favicon does not show in SERP


tpmccallum:
![Screenshot 2023-04-27 at 9 11 59 am](https://user-images.githubusercontent.com/9831342/234721940-08194c52-4f90-4d45-99f7-16a1143f23fa.png)

tpmccallum:
We have favicon images at the following location.
https://github.com/fermyon/developer/tree/main/static/image/icon

The site.toml files points to `favicon = "/static/image/favicon.png"` on line 12
https://github.com/fermyon/developer/blob/main/config/site.toml

I have created a new PR that updates the site.toml file to point into the `icon` directory.
https://github.com/fermyon/developer/pull/579

However, I am not sure if the above PR will make much difference as it seems like the favicon work is all being done by Handlebars i.e. https://github.com/fermyon/developer/blob/main/templates/content_top.hbs#L51

PR #579, opened 2023-04-26
Update location of favicon in site's config

tpmccallum:

Closes https://github.com/fermyon/developer/issues/578

I am a little perplexed because currently (without this change), the developer documentation correctly provides the URL to the favicon in the `head` of the HTML pages already.

Screenshot 2023-04-27 at 9 30 45 am

This update seems more correct than the `site.toml` linking to a folder other than `icon` (where the icons live). But any advice would be appreciated. @flynnduism do you have any thoughts on this?

### DIFF
--- a/config/site.toml
+++ b/config/site.toml
@@ -9,6 +9,6 @@ github = "https://github.com/fermyon/developer"
 twitter = "https://twitter.com/fermyontech"
 
 # Static assets that we need to provide
-favicon = "/static/image/favicon.png"
+favicon = "/static/image/icon/favicon.png"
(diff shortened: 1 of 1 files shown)


flynnduism:
/lgtm 

Cheers Tim. I believe it's defined in two places (the above one inside the `<head>`one in Bartholomew's [config/site.toml](https://github.com/fermyon/developer/blob/main/config/site.toml#L12)) - not sure which one overwrites the other, as long as it works! 😄

flynnduism (approved):

Issue #272, opened 2023-02-02
Ensure the cache directory (for Spin URL component sources) is well documented, so it's easy to go clean if needed.


tpmccallum:
Ensure the cache directory (for Spin URL component sources) is well documented, so it's easy to go clean if needed.
See https://github.com/fermyon/spin/pull/1088 for more details

tpmccallum:
PR at https://github.com/fermyon/developer/pull/587

tpmccallum:
Resolved at https://github.com/fermyon/developer/pull/587

PR #587, opened 2023-05-03
Document how to manage Spin cache

tpmccallum:


content/spin/managing-cache.md
```diff
@@ -0,0 +1,77 @@
+title = "Managing Cache"
```
itowlson, line 1: We have a bunch of caches - I'd strongly suggest clarifying this as the registry cache

content/spin/managing-cache.md
```diff
@@ -0,0 +1,77 @@
...
+| Windows | `{FOLDERID_LocalAppData}/spin/registry` | `C:\Users\Alice\AppData\Local/spin/registry` |
+
+For example inspecting the local `spin/registry` cache on macOS, (using the `tree` command) would look like this:
+
+<!-- @selectiveCpy -->
+
+```console
+$ tree ~/Library/Caches/spin/registry/
+
+├── data
```
itowlson, line 35: This is a phenomenal amount of detail to go into, and maybe a bit overwhelming?  I am not sure we would even guarantee the stability of this layout.  At this level of depth it feels more like "debugging Spin internals" kind of info rather than what an application developer would need to know.

content/spin/managing-cache.md
```diff
@@ -0,0 +1,77 @@
...
+
+> The `data` and `wasm` directories are content addressable. This means that if multiple applications reference the same static file or component source, Spin will be able to determine if it has already been pulled (on that users operating system), based on its digest. This also means that if an application has an update, Spin will only pull the changes in the component sources and static assets.
+
+The `manifests` directory contains the registry manifests for entire apps distributed with remote registries. They are placed in subdirectories that identify the application based on the registry, repository, and digest (or tag). The schema is aligned with the documentation on [distributing Spin apps using the OCI format](https://developer.fermyon.com/spin/distributing-apps).
+
+> When running an application from a remote registry, even if the application has already been pulled, Spin will first contact the registry to fetch the manifest.
+
+## Clearing Cached Files
```
itowlson, line 75: Honestly I think this is the main operation that a user will be interested in.

templates/spin_sidebar.hbs
```diff
@@ -64,6 +64,7 @@
             Tools
         </p>
         <ul class="menu-list">
+            <li><a {{#if (active_content request.spin-path-info "/spin/managing-cache" )}} class="active" {{/if}} href="{{site.info.base_url}}/spin/managing-cache">Managing Cache</a></li>
```
itowlson, line 67: Suggest in the Advanced section rather than Tools (if we do keep a separate page rather than merge onto the registry page).

itowlson (changes requested):
I'm a bit unclear who this is for - I can see developers wanting to clear the cache because it got big or to test cold start performance or whatever, but that feels like it could go in the existing registry documentation.  I feel it doesn't merit a whole page to itself (though if we decide it does I would definitely clarify that this is the _registry_ cache).

The details of the tree structure are more like design docs for someone maintaining or debugging the `spin-oci` crate.  I'm wary of suggesting to application developers that they need to know this stuff.  We _should_ have a home for this info, but it should maybe be in the Spin repository, maybe in an "as built" design docs area next to the SIPs?  Or a Spin Internals section here... but I'd argue for not mixed in with "how to install plugins and templates."

So... my suggestion would be to prune the content back to "registry stuff gets cached" and "how to clear the registry cache" (which includes the "where is the cache" bit), and move it onto the current registry page.  Others may have different views of course!

tpmccallum:
This closes https://github.com/fermyon/developer/issues/272

content/spin/managing-cache.md
tpmccallum: Thanks for the review @itowlson, great feedback.

Is this the only user-manageable cache? If so we can update the page name and the page content to use the more specific term "registry" for clarity. However, if there are other caches that are also user-manageable then perhaps we can leave the page name as it is and then separate each of the different caches into their own headings on this page. This content would go under its own heading and refer only to the registry cache.

templates/spin_sidebar.hbs
tpmccallum: The advanced section can work, yes!

By merging into the registry page, are we talking about tapping into a section like [this current documentation](https://developer.fermyon.com/spin/distributing-apps.md#running-published-applications-by-digest) so that we can mention the benefits of having a registry cache for the fetched resources? If so, that is a great idea (perhaps a brief note that mentions/links the advanced section as having more information on managing the registry cache).

content/spin/managing-cache.md
itowlson: There are a couple of caches in the plugins stuff, and the login token cache - not sure if there are others.  And some other stuff Spin puts on your system but we don't call it caches, e.g. templates.

I am not sure what you mean by "user manageable" in this context... the registry cache is not really "user manageable" except for you can delete it _grin_, and the plugins and login token caches have the same level of manageability.  With all of these, really, there's not much to say except "this is where it is, this is the impact if you delete it," but agree, if we are going to cover other caches, then a separate page makes sense... but I guess I am trying to get a clearer sense of the purpose... like is it tuning or tidying or "How to Obliterate Spin From Your System With Extreme Prejudice"?  I am kind of thinking what is the _task_ or _goal_ that we are trying to answer, if that makes sense...?

content/spin/managing-cache.md
tpmccallum: 100%
The other information is helpful for an outsider, though, to be fair. 
Caching is a very useful and practical feature, which leads to your next comment about it technically being an advanced feature. Especially diving into the details about cache being content addressable. Again, super efficient and robust but needs explaining (so users don't think the hash values are arbitrary or random). Personally, I think this needs its own page (and some mentions in other areas, as you have suggested).

content/spin/managing-cache.md
tpmccallum: Yes, this is a great point. The amount of page real estate this takes up is a bit much. I will trim it down but keep the layout because this is all by design. I think we need to find the middle ground between verbose and black box, right? :)

Let me trim it and we can see what it looks like.

content/spin/managing-cache.md
tpmccallum: Completely makes sense. How about a "Spin Cache" menu item in the "Advanced" section, and on that page, we explicitly refer to this PRs content as registry cache (which, as you say, it is)? 

The additional cache can go into that same "Advanced" area as its documentation is rolled out.

content/spin/managing-cache.md
tpmccallum: ... leaning towards how having its own H2 section would be sufficient (if there are other caches to talk about this one does not need its own whole page; they can share).

content/spin/managing-cache.md
tpmccallum: Please check the latest commit for updates.

content/spin/managing-cache.md
tpmccallum: Please check the latest commit for updates.

content/spin/managing-cache.md
tpmccallum: Please check the latest commit for updates.

templates/spin_sidebar.hbs
tpmccallum: Please check the latest commit for updates.

content/spin/cache.md
```diff
@@ -0,0 +1,64 @@
...
+
+> The `data` and `wasm` directories are content addressable. This means that if multiple applications reference the same static file or component source, Spin will be able to determine if it has already been pulled (on that users operating system), based on its digest. This also means that if an application has an update, Spin will only pull the changes in the component sources and static assets.
+
+The `manifests` directory contains the registry manifests for entire apps distributed with remote registries. They are placed in subdirectories that identify the application based on the registry, repository, and digest (or tag). The schema is aligned with the documentation on [distributing Spin apps using the OCI format](https://developer.fermyon.com/spin/distributing-apps).
+
+> When running an application from a remote registry, even if the application has already been pulled, Spin will first contact the registry to fetch the manifest.
+
+## Clearing Registry Cache Files
```
itowlson, line 62:
```suggestion
## Clearing the Registry Cache
```

reasoning: we don't want them thinking about individual files

content/spin/cache.md
```diff
@@ -0,0 +1,64 @@
...
+- [Viewing Registry Cache Files](#viewing-registry-cache-files)
+- [Clearing Registry Cache Files](#clearing-registry-cache-files)
+
+## The Spin Registry Cache
+
+A running Spin application can [fetch resources from remote registries](https://developer.fermyon.com/spin/spin-oci) and individual component sources via HTTP endpoints. These resources naturally consume network bandwidth. To ensure network efficiency, and to prevent waiting to download the same files every time an application is started, Spin automatically maintains a local `spin/registry` cache.
+
+## Viewing Registry Cache Files
+
+All downloadable application files that are automatically cached can be found in the following file paths, depending on the operating system being used to run the Spin application:
```
itowlson, line 20: Suggestion: before this paragraph, put a note or warning that the registry cache contains no user-serviceable parts; the only safe way to make changes to the contents of the cache is through ~an approved Electrolux dealer~ the Spin command line.  The only operation a user can safely undertake is to delete it.

content/spin/cache.md
itowlson, line 62: I would also move this ahead of "viewing files" because clearing is something users will want to do and looking inside is internals territory.

content/spin/cache.md
```diff
@@ -0,0 +1,64 @@
...
+
+- [The Spin Registry Cache](#the-spin-registry-cache)
+- [Viewing Registry Cache Files](#viewing-registry-cache-files)
+- [Clearing Registry Cache Files](#clearing-registry-cache-files)
+
+## The Spin Registry Cache
+
+A running Spin application can [fetch resources from remote registries](https://developer.fermyon.com/spin/spin-oci) and individual component sources via HTTP endpoints. These resources naturally consume network bandwidth. To ensure network efficiency, and to prevent waiting to download the same files every time an application is started, Spin automatically maintains a local `spin/registry` cache.
+
+## Viewing Registry Cache Files
```
itowlson, line 18: Not sure if you envisaging this covering other caches as the page title and some of the discussion implies; if so, this should be H3 so as to be subsidiary to the top-level "registry cache" topic.

content/spin/cache.md
```diff
@@ -0,0 +1,64 @@
+title = "Spin Cache"
```
itowlson, line 1: Suggest either "The Spin Registry Cache" or "Spin Caches" depending on how you see this page evolving.

content/spin/cache.md
```diff
@@ -0,0 +1,64 @@
...
+
+---
+
+- [The Spin Registry Cache](#the-spin-registry-cache)
+- [Viewing Registry Cache Files](#viewing-registry-cache-files)
+- [Clearing Registry Cache Files](#clearing-registry-cache-files)
+
+## The Spin Registry Cache
+
+A running Spin application can [fetch resources from remote registries](https://developer.fermyon.com/spin/spin-oci) and individual component sources via HTTP endpoints. These resources naturally consume network bandwidth. To ensure network efficiency, and to prevent waiting to download the same files every time an application is started, Spin automatically maintains a local `spin/registry` cache.
```
itowlson, line 16: Link goes to the tutorial rather than the detail material; is that intentional?  (The backlink is from the detail material.)

content/spin/cache.md
```diff
@@ -0,0 +1,64 @@
...
+
+The `manifests` directory contains the registry manifests for entire apps distributed with remote registries. They are placed in subdirectories that identify the application based on the registry, repository, and digest (or tag). The schema is aligned with the documentation on [distributing Spin apps using the OCI format](https://developer.fermyon.com/spin/distributing-apps).
+
+> When running an application from a remote registry, even if the application has already been pulled, Spin will first contact the registry to fetch the manifest.
+
+## Clearing Registry Cache Files
+
+Clearing the registry cache directory can be done by removing the `registry` directory entirely, or by removing the `registry`'s parent directory (`spin`). The only direct effect of these actions is that Spin will have to pull again all component sources and static assets.
```
itowlson, line 64: Removing the parent directory will clear other Spin caches too, which may have other effects.  Our advice should be to delete the `registry` directory.

itowlson, line 64:
```suggestion
Clearing the registry cache directory can be done by removing the `registry` directory entirely, or by removing the `registry`'s parent directory (`spin`). The only effect of these actions is that Spin will have to pull again all component sources and static assets.
```

reasoning: it sounded like we were sneakily avoiding mentioning the _indirect_ effects

content/spin/distributing-apps.md
```diff
@@ -80,6 +80,8 @@ $ spin up -f ghcr.io/alyssa-p-hacker/hello-world:v1
 
 > Remember that if the artifact is private you will need to be logged in, with permission to access it.
 
+Please note that Spin maintains a local [registry cache](https://developer.fermyon.com/spin/cache).
```
itowlson, line 83:
```suggestion
> Spin optimises downloads using a local [registry cache](https://developer.fermyon.com/spin/cache).
```

although the previous para is also a `>` note so maybe that won't look right

content/spin/cache.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+The `data` directory contains all static assets referenced from applications distributed with remote registries. The `wasm` directory contains all component sources referenced either in applications distributed with remote registries, or component sources from HTTP endpoints, directly referenced in `spin.toml` (for example, the static file server, or the KV explorer). 
+
+> The `data` and `wasm` directories are content addressable. This means that if multiple applications reference the same static file or component source, Spin will be able to determine if it has already been pulled (on that users operating system), based on its digest. This also means that if an application has an update, Spin will only pull the changes in the component sources and static assets.
+
+The `manifests` directory contains the registry manifests for entire apps distributed with remote registries. They are placed in subdirectories that identify the application based on the registry, repository, and digest (or tag). The schema is aligned with the documentation on [distributing Spin apps using the OCI format](https://developer.fermyon.com/spin/distributing-apps).
```
itowlson, line 58: I am not sure what "The schema is aligned with the documentation on distributing Spin apps using the OCI format" refers to - I don't think that page talks about the registry manifest schema.

content/spin/cache.md
```diff
@@ -0,0 +1,64 @@
...
+│           └── spin-openai-demo
+│               └── v1
+│                   ├── config.json
+│                   └── manifest.json
+└── wasm
+    ├── sha256:0b985e7d43e719f34cbb54849759a2f8e7913c0f9b17bf7cb2b3d2458d33859e
+    └── sha256:d5f9e1f6b61b90f7404e3800285f7860fe2cfc7d0116023efc370adbb403fe87
+```
+
+The `data` directory contains all static assets referenced from applications distributed with remote registries. The `wasm` directory contains all component sources referenced either in applications distributed with remote registries, or component sources from HTTP endpoints, directly referenced in `spin.toml` (for example, the static file server, or the KV explorer).
```
itowlson, line 54: Do we mention the KV explorer anywhere else?  Reading this para it feels like we expect the user to know what it is.

content/spin/cache.md
tpmccallum: Done, thanks!

content/spin/cache.md
tpmccallum: Done.

content/spin/cache.md
tpmccallum: Gone with The Spin Registry Cache.

content/spin/cache.md
tpmccallum: Yes, that was intentional.

content/spin/cache.md
tpmccallum: I will remove the last sentence and link. The main point of this page has already been made i.e. what is cache and how to clear it. This might be too much info. Thanks.

content/spin/cache.md
tpmccallum: Again, probably oversharing words. I will remove the content in the parenthesis.

content/spin/cache.md
```diff
@@ -0,0 +1,66 @@
...
+
+A running Spin application can [fetch resources from remote registries](https://developer.fermyon.com/spin/spin-oci) and individual component sources via HTTP endpoints. These resources naturally consume network bandwidth. To ensure network efficiency, and to prevent waiting to download the same files every time an application is started, Spin automatically maintains a local `spin/registry` cache.
+
+## Clearing the Registry Cache
+
+Clearing the registry cache directory can be done by removing the `registry` directory entirely. The only direct effect of these actions is that Spin will have to pull again all component sources and static assets.
+
+### Viewing Registry Cache Files
+
+> **Note:** The registry cache contains no user-serviceable parts; the only safe way to make changes to the contents of the cache is through the Spin command line. The only operation a user can safely undertake is to delete the `registry` directory.
```
itowlson, line 24: I need to be more careful about using informal, metaphorical language like "user-serviceable parts" in my comments, in case you like them, and I have to ask myself "would a non-native speaker recognise the idiom here."

I think it's okay for now (partly because right now I can't think of another good concise way to say it!) but we might need to iterate on a more literal way of expressing this.

itowlson (approved):
Thank you for your patience over the iterations of this one.  I left one comment about terminology which I acknowledge was originally my own; but regardless of that I reckon good to go.

tpmccallum:
Thanks for the review. Appreciate it.

PR #622, opened 2023-05-16
Accordion menus

karthik2804:
Adds accordion menu

image


flynnduism:
I have two comments -

* Collapse behavior (JS)
One when you open a section, it seems to collapse another.  I don't see a good reason for this as I think it makes it harder to find a page when you're browsing from section to section

*  Vague categories should be avoided (content)
I noticed a section called Advanced which has all sorts of different things mixed in.  This seems like it will make it hard to find the pages in here naturally. If things don't have a logical grouping, can we just ungroup them? (have a mix of accordion and non-accordion links)

karthik2804:
Happy to remove the animation and also restructure stuff as we see fit. We could definitely have a mixture.

itowlson:
Related to the auto collapse: this has the odd effect that if you are on a page in Section A, and expand Section B, your current page is no longer highlighted (or even visible) in the TOC.  So maybe another reason to remove the auto-collapse.

On the other hand I can see after someone has visited a few pages it could leave them with half the groups open and effectively back to a super long TOC.  Not sure what the compromise is.

itowlson:
@flynnduism The trouble with breaking the rag-bag of "advanced" topics up to top level is that it makes the most esoteric topics the most visible.  99.7% of users don't need to know how to author a template or how the registry cache is laid out: those shouldn't be the few pages that are always there asking to be clicked on.

We can for sure try to break out themes or move pages to other groups.  But there's a good chance we'll always be left with a few "look this is something we need to write down for the few niche users who care about it but it doesn't really relate to much else" kind of pages.  I'm not too uncomfortable with an "Advanced" or "Other Topics" kind of dumping area for those.  But yes we could always be more disciplined about using it!

itowlson:
Wow, I'm sorry @karthik2804, I went straight to the one thing that niggled me.  I really like this.  It looks good and definitely helps keep things tidy.  And that should have been my main message!

karthik2804:
@itowlson i appreciate the feedback. One potential solution is to have the currently active section always be open while closing the rest when browsing. It should be straightforward to implement that.

itowlson (approved):
Thanks for the update.  If @flynnduism is happy then I am happy.  We can keep iterating on this as we (and, more importantly, users) try it in practice.

tpmccallum (approved):

itowlson (approved):
Looks good!  My only niggle is that if I close the Get Started section (because I am now the SPIN MASTER) and then visit a new page, it pops open again.  But happy to iterate on that in post.

image

karthik2804:
Oooh, @itowlson nice catch! That is an easy fix. Fix on its way.

PR #627, opened 2023-05-17
Update Spin from 1.1 to 1.2

tpmccallum:

### DIFF
--- a/content/common/cli-reference.md
+++ b/content/common/cli-reference.md
@@ -105,6 +105,37 @@ SUBCOMMANDS:
 
 {{ blockEnd }}
 
+{{ startTab "v1.2.0"}}
+
+<!-- @selectiveCpy -->
+
+```console
+$ spin --help
+
+USAGE:
+    spin <SUBCOMMAND>
+
+OPTIONS:
+    -h, --help       Print help information
+    -V, --version    Print version information
+
+SUBCOMMANDS:
+    add          Scaffold a new component into an existing application
+    build        Build the Spin application
+    cloud        Commands for publishing applications to the Fermyon Platform
+    deploy       Package and upload an application to the Fermyon Platform
+    help         Print this message or the help of the given subcommand(s)
+    login        Log into the Fermyon Platform
+    new          Scaffold a new application based on a template
+    plugins      Install/uninstall Spin plugins
+    registry     Commands for working with OCI registries to distribute applications
+    templates    Commands for working with WebAssembly component templates
+    up           Start the Spin application
+    watch        Build and run the Spin application, rebuilding and restarting it when files
+```
+
+{{ blockEnd }}
+
 {{ blockEnd }}
 
 ### Add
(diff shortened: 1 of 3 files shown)


tpmccallum:
Closes https://github.com/fermyon/developer/pull/627

content/common/cli-reference.md
```diff
@@ -105,6 +105,37 @@ SUBCOMMANDS:
 
 {{ blockEnd }}
 
+{{ startTab "v1.2.0"}}
```
itowlson, line 108: The 1.0 tab is still in here; what is our policy on keeping docs for old versions around?

content/common/cli-reference.md
```diff
@@ -685,13 +955,48 @@ SUBCOMMANDS:
...
 
-{{ blockEnd }}
+{{ startTab "v1.2.0"}}
 
-> Please note: Spin `help` is a convenient way to access help using a subcommand, instead of using the `--help` option. For example, `spin help cloud` will give you the same output as `spin cloud --help`. Similarly, `spin help build` will give you the same output as `spin build --help` and so forth.
+<!-- @selectiveCpy -->
 
-### Login
+```console
+$ spin help
```
itowlson, line 963: This seems to entirely recapitulate `spin --help` from the top of the page.  Not sure why we would bother including it, especially when we haven't felt it was worth it in the past...?

itowlson (approved):

content/common/cli-reference.md
tpmccallum: We decided to strip out anything before v1.0.0 (we removed 0.x.0).
As far as moving forward, we might all need to weigh in on that. I don't have a preference other than we should probably only have a maximum of 5 tabs at any one time. I am even flexible on that. :)

itowlson:
Longer term, if we feel this needs to be in the docs, it would be great to find a way of presenting it that:

1. Makes it easier to review.  The current system means the entire help system output appears as inserted text; there's no sense of "what has changed" just a large mass of text to be checked, most of which has already been read over in several previous versions.
2. Makes it easier for readers to see what has changed across versions (though this could also be accomplished via a separate changelog).

I'm not sure what a good solution looks like though!  (And it's definitely not something we need to worry about for this PR.)

content/common/cli-reference.md
tpmccallum: It is a fair point. The main reason `spin help` is in the CLI Reference is because it is listed as a subcommand (a matter of just referencing what exists in the terminal). These docs are generated by stepping through each of the commands and their subcommands and `spin help` happens in that process. As an aside, I kinda like the indented note that explains how the `subcommand` `help` can be used in place of an `option` and how that relationship exists with other commands i.e. `spin help build` (is the same as `spin build --help`).

itowlson: Oh yeah the note is grand it was just the re-printing of the top level help.  But I misread that this was a new thing - we did the same on 1.0 and 1.1 so carry on.

_slinks back into curmudgeon cave_

content/common/cli-reference.md
```diff
@@ -2319,4 +3378,21 @@ CLI commands have four phases that indicate levels of stability:
...
+| ------------------------------------------------------------------------------------- | ------------ |
+| <code>spin add</code>                                                                 | Stable       |
+| <code>spin build</code>                                                               | Stable       |
+| <code>spin new</code>                                                                 | Stable       |
+| <code>spin plugins <install&vert;list&vert;uninstall&vert;update&vert;upgrade></code> | Stable       |
+| <code>spin templates <install&vert;list&vert;uninstall&vert;upgrade></code>           | Stable       |
+| <code>spin up</code>                                                                  | Stable       |
+| <code>spin cloud <deploy&vert;login></code>                                           | Stabilizing  |
+| <code>spin registry</code>                                                            | Stabilizing  |
+| <code>spin watch</code>                                                               | Experimental |
```
calebschoepp, line 3393: Probably stabilizing

tpmccallum: Thanks @calebschoepp  - new commit has this `watch` feature updated from Experimental to Stabalizing.

Issue #376, opened 2023-02-28
Lack of documentation for how to upgrade (Spin/Templates/Plugins/SDKs/AppConfig)


tpmccallum:
Those who have already used Spin to build an application may need to update and upgrade a variety of moving parts; at any stage in the future. For example:
- use the latest version of Spin
-use the latest version of a specific template
- use the latest version of a specific plugin
- update their application's configuration to point to a newer version of a Spin SDK etc.

Some things happen automatically and some have to be manually done, these tasks are not 100% clear.

tpmccallum:
Updating templates is at https://developer.fermyon.com/spin/managing-templates#upgrading-templates
Updating Plugins is at https://developer.fermyon.com/spin/managing-plugins#upgrading-plugins
We now need an overarching document that a) shows how to update Spin itself and b) links to these template and plugin docs; a one-stop-shop for a returning user who wants to, say, go from Spin 0.x.x to Spin 1.2.0.
Closing this because there is a fresh issue regarding exactly this task at https://github.com/fermyon/developer/issues/653

PR #639, opened 2023-05-18
Update install.md to ensure Spin is copied to correct path

tpmccallum:

### DIFF
--- a/content/spin/install.md
+++ b/content/spin/install.md
@@ -65,7 +65,7 @@ It's highly recommended to add Spin to a folder, which is on your path, e.g.:
 ```bash
 $ sudo mv spin /usr/local/bin/
 ```
-> **Hint:** If you already have Spin installed (i.e. via cargo) you can use `which spin` to check which folder to add the binary to. For example if `which spin` returns `/home/user/.cargo/bin` then you would `sudo mv spin /home/user/.cargo/bin` instead of the `/usr/local/bin` example above. Note: You can use `spin --version` to confirm the version.
+> If you have already installed Spin by building from source, and then install it via the installer, we recommend you remove the older source install by running `cargo uninstall spin-cli`  Otherwise the Cargo path may take precedence over the "install from binary" path, and commands may get the "wrong" version of Spin. Use `spin --version` to confirm the version on the PATH is the one you intend, or `which spin` to confirm the path it is found on.
 
 To install a specific version, you can pass arguments to the install script this way:
 
--- a/content/spin/install.md
+++ b/content/spin/install.md
@@ -34,7 +34,7 @@ It's highly recommended to add Spin to a folder, which is on your path, e.g.:
 ```bash
 $ sudo mv spin /usr/local/bin/
 ```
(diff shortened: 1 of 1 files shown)


karthik2804 (commented):
Not sure it is the recommended practice to move something downloaded to where cargo has installed a binary. A potential solution is to delete the cargo installation of Spin?

cc: @itowlson

tpmccallum:
Good point; I am not sure what best practice is in relation to removing things from `.cargo` area. Thanks for pointing out that we probably ought not just put binary files in there either. Would love a second opinion on how to proceed if a user used to use `cargo` and now want to use the install script.

content/spin/install.md
```diff
@@ -34,6 +34,7 @@ It's highly recommended to add Spin to a folder, which is on your path, e.g.:
 ```bash
 $ sudo mv spin /usr/local/bin/
 ```
+> **Hint:** If you already have Spin installed (i.e. via cargo) you can use `which spin` to check which folder to add the binary to. For example if `which spin` returns `/home/user/.cargo/bin` then you would `sudo mv spin /home/user/.cargo/bin` instead of the `/usr/local/bin` example above. Note: You can use `spin --version` to confirm the version.
```
itowlson, line 37: That directory belongs to Cargo; it would be preferable to discourage users from messing with it. 
 What about instead something like:

```suggestion
> If you have already installed Spin by building from source, and then install it via the installer, we recommend you remove the older source install by running `cargo uninstall spin-cli`  Otherwise the Cargo path may take precedence over the "install from binary" path, and commands may get the "wrong" version of Spin. Use `spin --version` to confirm the version on the PATH is the one you intend, or `which spin` to confirm the path it is found on.
```

content/spin/install.md
```diff
@@ -64,6 +65,7 @@ It's highly recommended to add Spin to a folder, which is on your path, e.g.:
...
```
itowlson, line 68: Same comment and suggestion.

itowlson (changes requested):
I feel really queasy encouraging users to move stuff that wasn't managed by Cargo into Cargo's private directory structure.  I've suggested uninstalling instead.

I'm not sure if you are also concerned about the "installed it from the script twice but moved it to two different directories" case.  The suggested new wording doesn't address that.  If we want to deal with that case as well, we'll need a separate paragraph, at which point we should maybe consider breaking it out into an "upgrading Spin" or "troubleshooting installation" section or it will clutter the main text.  (Actually maybe that is worth doing even with the current paragraph...)

content/spin/install.md
tpmccallum, line 68:
```suggestion
> If you have already installed Spin by building from source, and then install it via the installer, we recommend you remove the older source install by running `cargo uninstall spin-cli`  Otherwise the Cargo path may take precedence over the "install from binary" path, and commands may get the "wrong" version of Spin. Use `spin --version` to confirm the version on the PATH is the one you intend, or `which spin` to confirm the path it is found on.
```

tpmccallum: Thanks!

tpmccallum:
We have an issue in for explaining [how to update Spin](https://github.com/fermyon/developer/issues/376). This PR will cover off original point/issue of creating awareness of the spin version and path.

itowlson (approved):
LGTM - thanks!

PR #642, opened 2023-05-18
Update spin_sidebar.hbs

tpmccallum:

Using "Built with Spin" as this is shorter (would render well in the menu sidebar, as apposed to "Designing Spin Applications").

### DIFF
--- a/templates/spin_sidebar.hbs
+++ b/templates/spin_sidebar.hbs
@@ -13,6 +13,8 @@
                             href="{{site.info.base_url}}/spin/install">Install</a></li>
                     <li><a {{#if (active_content request.spin-path-info "/spin/quickstart" )}} class="active" {{/if}}
                             href="{{site.info.base_url}}/spin/quickstart">Quickstart</a></li>
+                    <li><a {{#if (active_content request.spin-path-info "/spin/designing-spin-apps.md" )}} class="active" {{/if}}
+                            href="{{site.info.base_url}}/spin/quickstart">Built With Spin</a></li>
                 </ul>
             </div>
             <div class="accordion-menu-item">
--- a/templates/spin_sidebar.hbs
+++ b/templates/spin_sidebar.hbs
@@ -13,7 +13,7 @@
                             href="{{site.info.base_url}}/spin/install">Install</a></li>
                     <li><a {{#if (active_content request.spin-path-info "/spin/quickstart" )}} class="active" {{/if}}
                             href="{{site.info.base_url}}/spin/quickstart">Quickstart</a></li>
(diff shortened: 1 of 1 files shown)


templates/spin_sidebar.hbs
```diff
@@ -13,6 +13,8 @@
                             href="{{site.info.base_url}}/spin/install">Install</a></li>
                     <li><a {{#if (active_content request.spin-path-info "/spin/quickstart" )}} class="active" {{/if}}
                             href="{{site.info.base_url}}/spin/quickstart">Quickstart</a></li>
+                    <li><a {{#if (active_content request.spin-path-info "/spin/designing-spin-apps.md" )}} class="active" {{/if}}
+                            href="{{site.info.base_url}}/spin/quickstart">Built With Spin</a></li>
```
karthik2804, line 17:
```suggestion
                            href="{{site.info.base_url}}/spin/designing-spin-apps">Built With Spin</a></li>
```

templates/spin_sidebar.hbs
```diff
@@ -13,6 +13,8 @@
                             href="{{site.info.base_url}}/spin/install">Install</a></li>
                     <li><a {{#if (active_content request.spin-path-info "/spin/quickstart" )}} class="active" {{/if}}
                             href="{{site.info.base_url}}/spin/quickstart">Quickstart</a></li>
+                    <li><a {{#if (active_content request.spin-path-info "/spin/designing-spin-apps.md" )}} class="active" {{/if}}
```
karthik2804, line 16:
```suggestion
                    <li><a {{#if (active_content request.spin-path-info "/spin/designing-spin-apps" )}} class="active" {{/if}}
```

karthik2804 (changes requested):

templates/spin_sidebar.hbs
itowlson, line 16: Don't put this in the Get Started section under its current title.  If we want to split the "showcase" / "see what other people are building" content from the "how to design larger applications" content (which is what the title reflects but does not currently exist) then we could put the former in Get Started.  The as-yet-unwritten app design stuff will belong under Applications though; it's definitely not starter material!

(I realise you've changed the title here but it needs to change on the page too, and I'd suggest changing the filename because we still want a "designing apps" page.)

itowlson (changes requested):

templates/spin_sidebar.hbs
tpmccallum: Closing this in favour of splitting the idea into two separate files. One to see what people are building with Spin and the other for explaining how to design and create large applications with Spin.

Issue #452, opened 2023-03-13
404 Page Enhancement


flynnduism:
Related to #451, @tpmccallum has requested updates to the 404 template

> Perhaps we could create a new GH task to make the 404 more visually appealing than just small text with links @flynnduism, any ideas on giving the 404 a makeover?

This is a nice-to-have, putting on the backlog here as seperate issue for now.

quyumkehinde:
This seems to be done. Closing the issue

PR #654, opened 2023-05-20
add sidebar when 404ing on particular project

karthik2804:
This adds the project sidebar when failing to find a page on particular projects

before
image

after
Screenshot 2023-05-20 at 12 14 40 PM

### DIFF
--- a/templates/main.hbs
+++ b/templates/main.hbs
@@ -5,16 +5,27 @@
     <div class="columns">
 
         {{! This adds the sidebar }}
-
+        {{#if (active_project request.spin-path-info "/spin/" )}}
+        {{> spin_sidebar }}
+        {{else}}
+        {{#if (active_project request.spin-path-info "/cloud/" )}}
+        {{> cloud_sidebar }}
+        {{else}}
+        {{#if (active_project request.spin-path-info "/bartholomew/" )}}
+        {{> bartholomew_sidebar }}
+        {{/if}}
+        {{/if}}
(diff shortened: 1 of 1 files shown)


itowlson (approved):
The "does not exist but has a TOC anyway" looks a bit odd to me; I wonder if there needs to be a corresponding update to the page body, to allude to the TOC if shown e.g.

---

### We couldn't find that topic.

* Try browsing {Spin|Cloud|...} topics on the left.
* Go to the Spin start page.
* Go to the Fermyon Cloud start page.
* Go to the home page to see all our documentation.

---

But that's not a blocker - happy to iterate on the wordsmithing in a follow up PR.

karthik2804:
If we can land on consensus on the wordsmithing. I can update the current PR too. I am happy either way. I do like the content you have suggested but not particularly sure how to put that in (Not sure if we wanted a bullet list on that page).

tpmccallum:
Thanks @karthik2804 and @itowlson 

After looking at this from a few different angles, it would be great to get this PR through first - alone.
The reason being when viewing from a mobile, the menu is a drop-down in the top right corner, and so wording like `left` etc. might be a little confusing. This is a great update. Will put some thought into the body of this page soon.

tpmccallum:
A 404 enhancement issue is underway at https://github.com/fermyon/developer/issues/452
Hopefully, we can improve the body by using images/art and links.

PR #657, opened 2023-05-22
fix css for project cards

karthik2804:


itowlson (approved):
I mean... I have absolutely no idea about CSS.

My approval is basically "I confirm that @karthik2804 is not trying to sneak in a Bitcoin miner," that's about all I can be sure of.

PR #660, opened 2023-05-23
Add border to images to distinguish between background and images

karthik2804:
Before 
image

After
image

Bonus content (light mode)
image

PS: @flynnduism is a CSS wizard!

### DIFF
--- a/static/css/styles.css
+++ b/static/css/styles.css
@@ -1749,6 +1749,15 @@ html.dark-theme .search-modal-container .result-section-container .suggested-pro
   background-color: #0000;
 }
 
+.documentation article.content-docs section#type img {
+  background: linear-gradient(50deg, #a87ce6 0%, #e6d2f1 100%);
+  border-radius: .75rem;
+  margin-top: 1.333em;
+  margin-bottom: 1.333em;
+  padding: .3rem;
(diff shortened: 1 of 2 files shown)


itowlson:
On what images does this appear?  (Like we wouldn't want it around the Fermyon logo for example.)

karthik2804:
Good catch - I will scope it down to be safer. It did not break the Fermyon logo but broke the project cards. I have scoped it down now to only have borders on images inside the body of doc pages.

tpmccallum:
This looks good.
Nice work.
One thing though, remember yesterday how the text colour on the Spin, Cloud and Bartholomew project boxes was all of a sudden incorrect? Same thing is happening again.

image

karthik2804:
That is because i did not create this branch from `main`. it should be fine once we merge to main.

karthik2804:
I can rebase if that would be preferred to be sure.

tpmccallum:
I guess we better check, so re base would be ideal if possible (so I can preview locally)

karthik2804:
@tpmccallum rebased!

tpmccallum:
Thanks, checking now.

tpmccallum (approved):
Looks really great, thanks @karthik2804

PR #675, opened 2023-06-05
Add a configuration variables tutorial

kate-goldenring:
Tutorial for new configuration variables feature.

References the cloud command reference page, so should be merged after https://github.com/fermyon/developer/pull/681

## Pre-Merge Content Checklist

This documentation has been checked to ensure that:

- [x ] The `title, `template`, and `date` are all set
- [x] Does this PR have a new menu item (anywhere in `templates/*.hbs` files) that points to a document `.md` that is set to publish in the future? If so please only publish the `.md` and `.hbs` changes in real-time (otherwise there will be a menu item pointing to a `.md` file that does not exist)
- [x] File does not use CRLF, but uses plain LF (hint: use `cat -ve <filename> | grep '^M' | wc -l` and expect 0 as a result) 
- [x] Has passed [`bart check`](https://developer.fermyon.com/bartholomew/quickstart)
- [x] Has been manually tested by running in Spin/Bartholomew (hint: use `PREVIEW_MODE=1` and run `npm run styles` to update styling)
- [x] Headings are using Title Case
- [x] Code blocks have the programming language set to properly highlight syntax and the proper copy directive
- [x] Have tested with `npm run test` and resolved all errors
- [ ] Relates to an existing (potentially outdated) blog article? If so please add URL in blog to point to this content.

PR #678, opened 2023-06-06
Install cloud plugin in install script

kate-goldenring:
I didn't update the other references to what the install script installs to say it also installs the cloud plugin. It didn't seem necessary to state since if not preinstalled it will be automatically installed later

DO NOT MERGE until after the 1.3 spin release

### DIFF
--- a/content/spin/install.md
+++ b/content/spin/install.md
@@ -166,7 +166,7 @@ This does not install any Spin templates or plugins. For a starter list, see the
 ## Installing Templates and Plugins
 
 Spin has a variety of templates and plugins to make it easier to create Spin applications in your favorite programming language. 
-The install script automatically installs a starter set of templates and plugins, namely templates from the Spin repository and JavaScript and Python toolchain plugins. If you used a different installation method, we recommend you install these templates and plugins manually, as follows:
+The install script automatically installs a starter set of templates and plugins, namely templates from the Spin repository and JavaScript and Python toolchain plugins and the Fermyon Cloud plugin. If you used a different installation method, we recommend you install these templates and plugins manually, as follows:
 
 ```sh
 spin templates install --git https://github.com/fermyon/spin --upgrade
@@ -175,6 +175,7 @@ spin templates install --git https://github.com/fermyon/spin-js-sdk --upgrade
 spin plugins update
 spin plugins install js2wasm --yes
 spin plugins install py2wasm --yes
+spin plugins install cloud --yes
 ```
 
 To list installed templates, run:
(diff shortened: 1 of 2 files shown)


content/spin/install.md
```diff
@@ -166,7 +166,7 @@ This does not install any Spin templates or plugins. For a starter list, see the
 ## Installing Templates and Plugins
 
 Spin has a variety of templates and plugins to make it easier to create Spin applications in your favorite programming language. 
-The install script automatically installs a starter set of templates and plugins, namely templates from the Spin repository and JavaScript and Python toolchain plugins. If you used a different installation method, we recommend you install these templates and plugins manually, as follows:
+The install script automatically installs a starter set of templates and plugins, namely templates from the Spin repository and JavaScript and Python toolchain plugins and the Fermyon Cloud plugin. If you used a different installation method, we recommend you install these templates and plugins manually, as follows:
```
itowlson, line 169: This is getting to the point where it might benefit from becoming a bullet list.

content/spin/install.md
```diff
@@ -175,6 +175,7 @@ spin templates install --git https://github.com/fermyon/spin-js-sdk --upgrade
 spin plugins update
 spin plugins install js2wasm --yes
 spin plugins install py2wasm --yes
+spin plugins install cloud --yes
```
itowlson, line 178: They don't actually need to do this though, do they?  Spin will automatically install the `cloud` plugin on demand?  (But maybe we prefer ahead-of-time install so they don't get a speed bump when they first log in.)

itowlson (approved):
Left a couple of comments but they're not blockers - LGTM!

content/spin/install.md
kate-goldenring: That was my thought process. It feels like it should really be built in rather than a "you're missing something" on the first install

itowlson: Sounds good - thanks!

content/spin/install.md
```diff
@@ -166,7 +166,7 @@ This does not install any Spin templates or plugins. For a starter list, see the
 ## Installing Templates and Plugins
```
macolso, line 166: Can we add a line that says users must be on Spin v1.3 or later to use this feature?

macolso:
We should also update the [Spin Limitations table in the FAQ documentation](https://developer.fermyon.com/cloud/faq#spin-limitations) to reflect that config is now supported!

macolso:
Whoops, you can ignore the above comments. I meant to leave them on #675

PR #681, opened 2023-06-07
Add Cloud Command Reference page

kate-goldenring:
Cloud commands

We will likely want to link to the cloud plugin blog from here. 

~Draft PR to see what folks think about having this added page in cloud to describe the `spin cloud` subcommands.~

~For some reason, the tabs are not rendering correctly :/~

~Note: Made up the existence of a cloud plugin `v0.1.1` that has an unhidden variables command. Likely we will actually replace this version~ removed

~TODO: add definition of what supported spin version for each cloud plugin version~

### DIFF
--- a/content/cloud/cloud-plugin.md
+++ b/content/cloud/cloud-plugin.md
@@ -17,7 +17,7 @@ url = "https://github.com/fermyon/developer/blob/main/content/cloud/cloud-plugin
 
 ## Spin Cloud Command
 
-Fermyon provides a `cloud` plugin for the Spin CLI for you to manage Spin applications in Fermyon Cloud. This page documents the `spin cloud` command. Specifically, all of the available options and subcommands. For more information on subcommand stability, see the [subcommands stability table](#subcommand-stability-table). You can reproduce the Spin Cloud command documentation on your machine by using the --help flag. For example:
+Fermyon provides a []`cloud` plugin](https://github.com/fermyon/cloud-plugin) for the Spin CLI for you to manage Spin applications in Fermyon Cloud. This page documents the `spin cloud` command. Specifically, all of the available options and subcommands. For more information on subcommand stability, see the [subcommands stability table](#subcommand-stability-table). You can reproduce the Spin Cloud command documentation on your machine by using the `--help` flag. For example:
 
 {{ tabs "cloud-plugin-version" }}
 
--- a/content/cloud/cloud-plugin.md
+++ b/content/cloud/cloud-plugin.md
@@ -16,11 +16,12 @@ url = "https://github.com/fermyon/developer/blob/main/content/cloud/cloud-plugin
   - [Subcommands Stability Table](#subcommands-stability-table)
 
 ## Spin Cloud Command
+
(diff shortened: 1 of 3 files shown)


itowlson:
I think the tabs issue is they need `enable_shortcodes` in the page header.  (Not sure why this isn't the default.  I always forget it.)

itowlson:
This structure seems good to me.  In the sidebar, I'd suggest renaming "CLI Reference" to "Spin CLI Reference" and moving "Cloud Command Reference" up one to be next to it.  But that's just minor tweaks.

karthik2804:
@itowlson It is better not to have `enable_shortcodes` by default because it comes with an added cost to rendering.

content/cloud/cloud-plugin.md
```diff
@@ -0,0 +1,345 @@
+title = "Cloud Command Reference"
+date = "2023-06-07T22:41:02.478752Z"
+template = "cloud_main"
```
karthik2804, line 3:
```suggestion
template = "cloud_main"
enable_shortcodes = true
```

kate-goldenring:
@karthik2804 are the tabs okay with `>=` sign? They seem to be rendering fine

kate-goldenring:
Should we link to the cloud-plugin repo?

karthik2804:
@kate-goldenring they render find and will consistently match across tabs if they are the same string. what would not work is if using something like `>=1.2` and then having another like `>=1.1`, `1.2`.
The above example would not break per say but auto switching of tabs would not work as expected.

kate-goldenring:
@karthik2804 @itowlson @melissaklein24 I think this should be ready to review. I added the note about spin compatibility within each version tab instead of in the tab name. Also, instead of replacing the 0.1.0 plugin, we are putting out a 0.1.1 plugin so added tabs for it

content/cloud/cloud-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+  - [Login](#login)
+  - [Variables](#variables)
+    - [Set (Variables)](#set-variables)
+    - [Delete (Variables)](#delete-variables)
+    - [List (Variables)](#list-variables)
+  - [Subcommands Stability Table](#subcommands-stability-table)
+
+## Spin Cloud Command
+
+Fermyon provides a []`cloud` plugin](https://github.com/fermyon/cloud-plugin) for the Spin CLI for you to manage Spin applications in Fermyon Cloud. This page documents the `spin cloud` command. Specifically, all of the available options and subcommands. For more information on subcommand stability, see the [subcommands stability table](#subcommand-stability-table). You can reproduce the Spin Cloud command documentation on your machine by using the `--help` flag. For example:
```
itowlson, line 20: Is the extra `]` just before `cloud` (at the start of the para) intentional?

content/cloud/cloud-plugin.md
```diff
@@ -0,0 +1,363 @@
...
+
+{{ tabs "cloud-plugin-version" }}
+
+{{ startTab "v0.1.0"}}
```
itowlson: As 0.1.1 will be the 'current' version at Spin 1.3 (which is itself the first time users can run `cloud` as a plugin), I'd suggest dropping the 0.1.0 tab.  Not a blocker if other folks want to keep it though!

kate-goldenring:
>Also, instead of replacing the 0.1.0 plugin with unhidden variables commands, we are putting out a 0.1.1 plugin so added tabs for it

Chatted with @itowlson and decided to remove `v0.1.0` per his recommedation rather than add docs we dont assume anyone will use

itowlson (approved):
LGTM - one non-blocking suggestion (and one possible typo where there's no need to re-review if it does need a fix)

PR #683, opened 2023-06-08
fix jank on multitab on codeblock load

karthik2804:
fixes layout shift on page load when initializing the multitab blocks.


itowlson (approved):
Soooooooooo much better.  Nice one @karthik2804

PR #684, opened 2023-06-08
fix broken links

karthik2804:


content/spin/contributing-docs.md
```diff
@@ -70,7 +70,18 @@ The resulting output would be as follows.
...
+```toml
+.
+.
+body_source = "<path to the content>"
+
+[extra]
+
+```
+
+The value for the key `body_source` should be the path from which the content should be shared relative to the `content` folder.
```
itowlson, line 84: An example might help (though not super necessary).

itowlson (approved):
Comment is non-blocking

content/spin/contributing-docs.md
karthik2804: I just wanted to add something in there to fix links. I can add an example in a future PR once the documentation for Bartholomew is added for it.

PR #698, opened 2023-06-20
Adding 1.3 to the cli reference

melissaklein24:

- [X] Has passed [`bart check`](https://developer.fermyon.com/bartholomew/quickstart)
- [X] Have tested with `npm run test` and resolved all errors

### DIFF
--- a/content/spin/cli-reference.md
+++ b/content/spin/cli-reference.md
@@ -6,170 +6,43 @@ enable_shortcodes = true
 url = "https://github.com/fermyon/developer/blob/main/content/spin/cli-reference.md"
 
 ---
-- [Spin](#spin)
-  - [Add](#add)
-  - [Build](#build)
-  - [Cloud](#cloud)
-    - [Deploy (Cloud)](#deploy-cloud)
-    - [Login (Cloud)](#login-cloud)
-  - [Deploy](#deploy)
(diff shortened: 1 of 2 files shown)


content/spin/cli-reference.md
```diff
@@ -410,6 +508,31 @@ SUBCOMMANDS:
 
 {{ blockEnd }}
 
+{{ startTab "v1.3.0"}}
+
+<!-- @selectiveCpy -->
+
+```console
```
itowlson, line 515: All of this going to depend on what version of the cloud plugin they have installed.  It might be better to link to https://developer.fermyon.com/cloud/cloud-plugin which covers the cloud commands with appropriate versioning.  But I am not sure what the marketing position is, if we want to include cloud commands in the Spin docs.

itowlson (changes requested):
I'm afraid I never have the endurance to read these in detail, but I do think that now cloud is releasing with its own versioning and its own cadence, we need to break those commands out to a separate section or page (as the Cloud docs now do).  Whether that should only be in the Cloud docs, or whether we should copy it to the Spin docs too, I will leave to ~the mighty gods on high Olympus~ marketing to decide.

mikkelhegn:
Closing in favor of #702, which included these changes.

PR #756, opened 2023-07-24
add script to enable publish by time

karthik2804:
Content can be wrapped around with the following tag in templates to enable displaying them only once a preset time is past. 

 ```
{{#if (timed_publish "2023-07-25T17:26:00Z" env.PREVIEW_MODE)}}

// Content

{{/if}}
```

### DIFF
--- a/scripts/timed_publish.rhai
+++ b/scripts/timed_publish.rhai
@@ -1,7 +1,7 @@
 let time = params[0];
 let preview = params[1];
 
-console_println("this si the value");
+console_println("This is the value");
 console_println(preview);
 
 if (preview == "1") {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: scripts/timed_publish.rhai [hunk 1]
```diff
@@ -0,0 +1,15 @@
+let time = params[0];
+let preview = params[1];
+
+console_println("this si the value");
```
tpmccallum, line 4:
```suggestion
console_println("This is the value");
```

tpmccallum (commented):
Just going to test this locally and will approve soon.

tpmccallum (approved):

tpmccallum:
Thanks @karthik2804 
This works great (even nesting the publish time `if` inside a sidebar menu's `if`.
Thanks again; this will come in handy in the future.

PR #778, opened 2023-08-14
Fix path reference in contrib guide for Spin Up Hub

mikkelhegn:
Small correction to where to find the Spin Up Hub files.

### DIFF
--- a/content/api/hub/contributing.md
+++ b/content/api/hub/contributing.md
@@ -77,7 +77,7 @@ $ git remote add upstream https://github.com/fermyon/developer
 
 ### Creating Markdown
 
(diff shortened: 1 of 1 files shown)


karthik2804 (approved):

mikkelhegn:
@karthik2804 What should I do with the failing test? Also failing here #780 and here #779

karthik2804:
Looks like this is a broken link  https://docs.sigstore.dev/cosign/installation/. It is not related to this PR but it is broken on main I think. 

The `https://www.godaddy.com` link is a 403 and I tested that it fails with curl locally but the website works normally so we need to add it to our exceptions list. 
https://github.com/fermyon/developer/blob/c5700c969353479f7440c77acc35931f8e09a781/.build/check-broken-links.sh#L75

cc: @mikkelhegn

mikkelhegn:
Created an issue based on the above. I'll go ahead and merge these PR's.

Issue #788, opened 2023-08-17
JSON-stringifying a JavaScript error gives an empty object


macolso:
Would be good to add this callout to JS API guide for SQLite: https://developer.fermyon.com/spin/sqlite-api-guide

itowlson:
@macolso I renamed the issue to reflect the user impact, and that (from what @karthik2804 has indicated to me) this is part of the JavaScript error standard and is not specific to Spin SQLite errors.  Feel free to revert or further edit the title if you don't like it!

This does kind of raise a question about how much it is our job to document language/runtime surprises.  I can reproduce the exact same behaviour in plain ol' Node and plain ol' JS.  If the Spin docs take on the job of warning users about all the surprises in JavaScript then we could be here for a while...!

tpmccallum:
Good point @itowlson 
However, we could add just a small note on the last line of the general notes in the Typescript tab (see image below).
All of the error I induce are actually surfaced as exceptions both in the Spin CLI and returned in the response.
What code can you suggest to break the following `index.js` so I can see the empty object (in place of a real error)?

```javascript
import {Sqlite} from "@fermyon/spin-sdk"
const encoder = new TextEncoder()

export async function handleRequest(request) {
    const conn = Sqlite.openDefault();
    const result = conn.execute("SELECT * FROM character;");
    const json = JSON.stringify(result.rows);
    
    return {
        status: 200,
        headers: { "foo": "bar" },
        body: encoder.encode(json).buffer
    }
}
```
![Screenshot 2023-08-18 at 11 59 21](https://github.com/fermyon/developer/assets/9831342/1580c067-eef1-4c96-863c-3ec146302fc7)

itowlson:
@tpmccallum To see the issue, add this anywhere before the final `return`: 

```
try {
  throw new Error("oh no");
} catch (err) {
  const errjson = JSON.stringify(err);
  console.log(errjson);
  return { status: 200, headers: {}, body: encoder.encode(errjson).buffer };
}
```

itowlson:
Or if you want to see the same effect being generated from the Spin SDK, change the SQL statement to `SELECT * FROM characterzzzzzzzzzzzz` and put it in a try-catch.

(Without the try-catch the unhandled error will be stringised and returned in the HTTP response, which is what you're seeing when you see error responses.  To see the JSON problem you have to `catch` the exception and `JSON.stringify` that.)

[...]

PR #791, opened 2023-08-18
Add exception handling sentence.

tpmccallum:

closes https://github.com/fermyon/developer/issues/788


itowlson (changes requested):
Reminder that whatever wording we adopt will need to be propagated across all Spin API guides that contain fallible functions.

itowlson (changes requested):
The wording is okay (okay I would say that) but this change applies to _all_ fallible JS APIs.  It's not specific to SQLite.

itowlson (approved):

Issue #820, opened 2023-08-24
Hub cards can have very little text


mikkelhegn:
The top left one have the title and tags overlap
<img width="1231" alt="image" src="https://github.com/fermyon/developer/assets/12242001/e1c0c25f-0de3-453d-935c-8498285da7ed">

Could consider re-design to make room for more, and/or check the length in the PR contributions.

tpmccallum:
Hi @flynnduism and @karthik2804 
This could be solved in one or more of the following ways:

1. Limit the amount of text allowed in a title
2. Reduce the font of title
3. Increase the layout of the card

I am not sure which one (perhaps a combination of 2 and 3 might be the way to go)

PR #821, opened 2023-08-24
Shortening title for better layout

mikkelhegn:
#820 is probably the real fix.


tpmccallum (approved):

tpmccallum:
Thanks @mikkelhegn 
I see there is another issue to address the length of title and/or size of the layout and/or the size of the font (so longer titles can fit). This PR at least makes it render neatly for now.
Thanks

PR #839, opened 2023-08-28
move building hub to ci

### LINKED ISSUE #825: [Discussion]: Should the compiled VueJS application be removed from source control
Having the compile representation of the Vue application for the Hub in source control, makes keeping multiple PRs independent really hard and requires the reviewer to run `npm run build` in the `spin-up-hub` sub-folder.

IMO building the Vue application for the hub must happen in parallel to running `spin watch` in the root folder locally, and obviously CI/CD must build the app too before publishing a new version to the webserver

karthik2804:
Deleting the `dist` folder in the spin-up hub in favor of building it in CI to avoid conflicts as well as the index folder for the content of the hub. 

closes #825


karthik2804:
Something seems to be off with `spin watch` here where the server keeps restarting - Have to investigate what is going on.

karthik2804:
Known issue documented at https://github.com/fermyon/spin/issues/1692

vdice:
Looks good to me, but we may have to update the `npm test` logic to avoid hitting https://github.com/fermyon/spin/issues/1692.  Currently, that behavior prevents the server from running and thus the test job fails.  Perhaps we change https://github.com/fermyon/developer/blob/main/.build/check-broken-links.sh#L53 to just be `spin up`?

karthik2804:
@vdice - The problem with that would be that it would affect local development as well. WE would have to move to using `spin up` or move back to `nodemon` until the above is fixed.

vdice:
@karthik2804 good point.  Defer to you... `nodemon` I suppose, to retain "watch" parity?

karthik2804:
@vdice - I would slightly lean towards inserting `nodemon` as a dependency, for now, to get this through if that sounds like a plan.

vdice:
Sure.  We can track the issue and go back to `spin watch` when ready.

karthik2804:
There were some broken links so I have fixed them as well. @vdice this should be good for a review with the updates to nodemon and fixing the broken links.

package.json
```diff
@@ -37,8 +37,8 @@
     "yaml-front-matter": "^4.1.1"
   },
   "scripts": {
-    "spin": "spin watch --file spin.toml --quiet --env PREVIEW_MODE=$PREVIEW_MODE -e BASE_URL=http://localhost:3000",
-    "bundle-scripts": "parcel build static/js/src/main.js --dist-dir static/js --no-source-maps",
+    "spin": "nodemon --watch content --watch static --watch templates --watch spin-up-hub --ext html,md,rhai,hbs,css,js --verbose --legacy-watch --signal SIGINT --exec 'npm run build-index && spin up --file spin.toml --quiet --env PREVIEW_MODE=$PREVIEW_MODE'",
+    "bundle-scripts": "parcel watch static/js/src/main.js --dist-dir static/js --no-source-maps",
```
vdice, line 41: Double-checking: was this meant to change to `watch` here as well?

vdice (commented):
Prior to merge, let's poke around a little bit more to definitively state the restart behavior is due to https://github.com/fermyon/spin/issues/1692...

package.json
karthik2804: The `watch` on bundle-scripts was an artifact of my testing the feedback changes. It needs to be `build` when we eventually want to move the building of CSS and JS to the CI/CD pipeline as well.  This is because `parcel watch` by default looks for changes in its input files and never terminates. 
we might just add new npm script commands for when that happens.

package.json
```diff
@@ -37,8 +37,8 @@
     "yaml-front-matter": "^4.1.1"
   },
   "scripts": {
-    "spin": "spin watch --file spin.toml --quiet --env PREVIEW_MODE=$PREVIEW_MODE -e BASE_URL=http://localhost:3000",
-    "bundle-scripts": "parcel build static/js/src/main.js --dist-dir static/js --no-source-maps",
+    "spin": "nodemon --watch content --watch static --watch templates --watch spin-up-hub --ext html,md,rhai,hbs,css,js --verbose --legacy-watch --signal SIGINT --exec 'npm run build-index && spin up --file spin.toml --quiet --env PREVIEW_MODE=$PREVIEW_MODE'",
```
vdice, line 40:
```suggestion
    "spin": "nodemon --watch content --watch static --watch templates --watch spin-up-hub --ext html,md,rhai,hbs,css,js --verbose --legacy-watch --signal SIGINT --exec 'npm run build-index && npm run build-hub-index && spin up --file spin.toml --quiet --env PREVIEW_MODE=$PREVIEW_MODE'",
```

itowlson:
In reference to https://github.com/fermyon/developer/pull/839#issuecomment-1694807736 the issue seems to be that the `bartholomew` component requests a rebuild whenever `static` (among other directories) is changed, but its build command updates the indexes, which are in `static` so... infinite loop.

As the `bartholomew` build command appears to only be interested in `content` and `content/api/hub`, I am not sure why `static` changes merit a rebuild... can this (and other?) directories not just be taken out of the `bartholomew` `build,watch` list?

PR #843, opened 2023-08-31
Trigger index rebuild only when content changes

itowlson:
The current `build.watch` configuration:

* Rebuilds the index files in `static`
* Depends on multiple directories including `static`

This causes `spin watch` to go into an infinite loop as the build updates the index files in static, which triggers a rebuild, which updates the index files in static, which...

The only actual inputs to the indexes are under `content`, so this is the only directory we need to watch _for rebuild_.

(Yes, other directories require a reload, i.e. restart `spin up`.  That will still happen because `spin watch` considers the `files`.  But these don't require a re-run of the index build commands.)

### DIFF
--- a/spin.toml
+++ b/spin.toml
@@ -14,7 +14,7 @@ files = [ "content/**/*" , "templates/*", "scripts/*", "config/*", "shortcodes/*
 route = "/..."
 [component.build]
 command = "npm run build-index && npm run build-hub-index"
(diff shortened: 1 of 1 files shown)


vdice (approved):
Aha, thank you @itowlson !

tpmccallum:
@vdice with this one now merged, is a restart of the `spin watch ...` itself going to take place (so that the `spin.toml` file's new changes will catch on), or can we do a manual restart as a once-off task, for this PR?
Thanks
Tim M.

PR #863, opened 2023-09-05
fix(install): Update install script to install canary plugins

kate-goldenring:
Installs canary plugins when flag is specified and warns that templates must be manually modified to target canary SDKs.

~Note: python SDK canary is currently behind latest. We should amend that before we merge this.~ The canary artifacts are updated despite the canary release looking old

### DIFF
--- a/downloads/install.sh
+++ b/downloads/install.sh
@@ -4,6 +4,7 @@ set -euo pipefail
 # Fancy colors for output
 RED='\033[0;31m'
 GREEN='\033[0;32m'
+YELLOW='\033[0;33m'
 NC='\033[0m' # No Color aka reset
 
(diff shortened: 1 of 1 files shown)


tpmccallum (approved):

downloads/install.sh
```diff
@@ -131,13 +134,22 @@ fancy_print 0 "Step 4: Install default templates"
 ./spin templates install --git "https://github.com/fermyon/spin-python-sdk" --upgrade
 ./spin templates install --git "https://github.com/fermyon/spin-js-sdk" --upgrade
```
itowlson, line 135: Might be good to install the templates from main too?

Issue #748, opened 2023-07-19
curl installer errors on step 2


flynnduism:
@taylols and I hit this error when installing Spin 1.4 via the doc's curl installer:`curl -fsSL https://developer.fermyon.com/downloads/install.sh | bash`

**Error message:**

```
Step 1: Downloading: https://github.com/fermyon/spin/releases/download/v1.4.1/spin-v1.4.1-macos-aarch64.tar.gz
Done...

Step 2: Decompressing: spin-v1.4.1-macos-aarch64.tar.gz
x crt.pem
x spin.sig
x README.md
x LICENSE
x spin: Can't replace existing directory with non-directory
tar: Error exit delayed from previous errors.
```

The error was confusing - there was no previous `spin-v1.4.1-macos-aarch64.tar.gz` in the directory in question so this conflict was hard to understand.

**Workaround:** 

* deleted the downloaded `tar.gz` and extracted files (to ensure no conflict) and tried again - ❌ same error
* create a new subdirectory, run the curl installer inside there ✅ it works!

**Suggestions:**

* Not sure what prompted the directory error, but if it's common perhaps the Install docs could include a `mkdir spin && cd spin` precursor step?
* Or save that kind of instruction to be contextual in response to the installer CLI exiting with `Can't replace existing directory with non-directory`?
* This kind of thing could be documented under our [Troubleshooting](https://developer.fermyon.com/spin/troubleshooting-application-dev) section - searching w/ the error message provided no answers

tpmccallum:
I finally got this same message today @flynnduism and @itowlson 
Essentially, the message arises when there is a `spin` folder in the working path that `curl -fsSL https://developer.fermyon.com/downloads/install.sh | bash` is being run. For example:

```
$ curl -fsSL https://developer.fermyon.com/downloads/install.sh | bash
Step 1: Downloading: https://github.com/fermyon/spin/releases/download/v1.5.0/spin-v1.5.0-macos-amd64.tar.gz
Done...

Step 2: Decompressing: spin-v1.5.0-macos-amd64.tar.gz
x crt.pem
x spin.sig
x README.md
x LICENSE
x spin: Can't replace existing directory with non-directory
tar: Error exit delayed from previous errors.

```

Decision, should we:
-  [ ] add a message in the docs for users to remove the `spin` directory first (before using the installer)
-  [ ] update the error message in the Spin CLI (which means that we need to move this issue back over to Spin project itself)
cc: @mikkelhegn

[...]

PR #889, opened 2023-09-19
Add a check to see if a `spin` item is in the way of the installer succeeding (before we even download the spin tarball)

tpmccallum:

I was finally able to reproduce an error that has been flagged in an issue. This update closes https://github.com/fermyon/developer/issues/748 by adding a new step in the installer, which checks to see if a `spin` item is in the way of the installer succeeding (before the user’s end expends any resources on downloading the tarball).

### DIFF
--- a/downloads/install.sh
+++ b/downloads/install.sh
@@ -116,7 +116,7 @@ URL="https://github.com/fermyon/spin/releases/download/${VERSION}/${FILE}"
 current_dir=$(pwd)
 
 # Define a list of items to check for (this can be updated if the spin tarball is updated)
-items=("/spin" "/LICENSE" "/crt.pem" "spin.sig", "/README.md")
+items=("/spin")
 
 # Create a variable that can be called on to question if an exit 1 is required (due to items already existing)
 exit_due_to_existing_item=0
--- a/downloads/install.sh
+++ b/downloads/install.sh
@@ -114,7 +114,6 @@ URL="https://github.com/fermyon/spin/releases/download/${VERSION}/${FILE}"
 
 # Establish the location of current working environment
 current_dir=$(pwd)
-fancy_print 0 "Current working directory: ${current_dir}"
 
 # Establish the absolute Spin installation location
 spin_installation_location=${current_dir}"/spin"
--- a/downloads/install.sh
+++ b/downloads/install.sh
@@ -123,7 +123,7 @@ exit_due_to_existing_item=0
 
 # Iterate through the list of items
 for item in "${items[@]}"; do
-    if [ -e "${current_dir}${item}" ]; then
+    if [ -d "${current_dir}${item}" ]; then
         # Update the variable because the found existing item now warrants an exit 1 after this loop
         exit_due_to_existing_item=1
         fancy_print 1 "Error ${item} already exists, please delete ${current_dir}${item} and run the installer again."
--- a/downloads/install.sh
+++ b/downloads/install.sh
@@ -126,7 +126,7 @@ for item in "${items[@]}"; do
     if [ -d "${current_dir}${item}" ]; then
         # Update the variable because the found existing item now warrants an exit 1 after this loop
         exit_due_to_existing_item=1
-        fancy_print 1 "Error ${item} already exists, please delete ${current_dir}${item} and run the installer again."
+        fancy_print 1 "Error: ${item} already exists, please delete ${current_dir}${item} and run the installer again."
     fi
 done
 
--- a/downloads/install.sh
+++ b/downloads/install.sh
@@ -115,23 +115,11 @@ URL="https://github.com/fermyon/spin/releases/download/${VERSION}/${FILE}"
 # Establish the location of current working environment
 current_dir=$(pwd)
 
-# Define a list of items to check for (this can be updated if the spin tarball is updated)
-items=("/spin")
+# Define Spin directory name
+spin_directory_name=("/spin")
 
-# Create a variable that can be called on to question if an exit 1 is required (due to items already existing)
-exit_due_to_existing_item=0
-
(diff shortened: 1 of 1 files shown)


downloads/install.sh
```diff
@@ -112,6 +112,18 @@ fi
...
 
+# Establish the location of current working environment
+current_dir=$(pwd)
+fancy_print 0 "Current working directory: ${current_dir}"
+
+# Establish the absolute Spin installation location
+spin_installation_location=${current_dir}"/spin"
+
+# Check to see if an item called spin already exists (before expending resources downloading the file).
+if [ -e "$spin_installation_location" ]; then
```
itowlson, line 123: I believe this is only an issue if it's a directory - if existing files are a problem then we should also check for `LICENSE`, `README.md` and the other files in the tarball.  (I mean... we could check if there are directories with those names too... but that seems a lot less likely than there being files.)

downloads/install.sh
```diff
@@ -112,6 +112,18 @@ fi
...
+    fancy_print 1 "Error ${spin_installation_location} already exists, please delete ${spin_installation_location} and run the installer again."; exit 1
```
itowlson, line 124:
```suggestion
    fancy_print 1 "Error: ${spin_installation_location} already exists, please delete ${spin_installation_location} and run the installer again."; exit 1
```

downloads/install.sh
```diff
@@ -112,6 +112,18 @@ fi
 FILE="spin-${VERSION}-${OS}-${ARC}.tar.gz"
 URL="https://github.com/fermyon/spin/releases/download/${VERSION}/${FILE}"
 
+# Establish the location of current working environment
+current_dir=$(pwd)
+fancy_print 0 "Current working directory: ${current_dir}"
```
itowlson, line 117: I'd suggest not printing this except when testing/debugging

tpmccallum, line 117:
```suggestion
```

downloads/install.sh
tpmccallum: Good point, indeed. I have updated in the next push.

tpmccallum: The script now checks for the additional items (not just `spin`), we are using `items` as not to expend additional syntax discerning if things are files or directories. If it exists, then the installer will exit.

tpmccallum:
If any items exist (that should not be there), it lists all of them in a group once, so the user knows what to do (entirely) before trying the installer again. 
The exit is conditional (based on if one or more items (that should not be there).
Screenshot 2023-09-19 at 12 47 23

downloads/install.sh
```diff
@@ -112,6 +112,29 @@ fi
...
+
+# Define a list of items to check for (this can be updated if the spin tarball is updated)
+items=("/spin" "/LICENSE" "/crt.pem" "spin.sig", "/README.md")
+
+# Create a variable that can be called on to question if an exit 1 is required (due to items already existing)
+exit_due_to_existing_item=0
+
+# Iterate through the list of items
+for item in "${items[@]}"; do
+    if [ -e "${current_dir}${item}" ]; then
```
itowlson, line 126:
```suggestion
    if [ -d "${current_dir}${item}" ]; then
```

downloads/install.sh
```diff
@@ -112,6 +112,29 @@ fi
...
+        # Update the variable because the found existing item now warrants an exit 1 after this loop
+        exit_due_to_existing_item=1
+        fancy_print 1 "Error ${item} already exists, please delete ${current_dir}${item} and run the installer again."
```
itowlson, line 129:
```suggestion
        fancy_print 1 "Error: ${item} already exists, please delete ${current_dir}${item} and run the installer again."
```

karthik2804, line 129: Based on the output - it seems to  start with `/` instead of `./`.

```suggestion
        fancy_print 1 "Error .${item} already exists, please delete ${current_dir}${item} and run the installer again."
```

downloads/install.sh
```diff
@@ -112,6 +112,29 @@ fi
 FILE="spin-${VERSION}-${OS}-${ARC}.tar.gz"
 URL="https://github.com/fermyon/spin/releases/download/${VERSION}/${FILE}"
 
+# Establish the location of current working environment
+current_dir=$(pwd)
+
+# Define a list of items to check for (this can be updated if the spin tarball is updated)
+items=("/spin" "/LICENSE" "/crt.pem" "spin.sig", "/README.md")
```
tpmccallum, line 119:
```suggestion
items=("/spin")
```

downloads/install.sh
tpmccallum: Nice catch @karthik2804 I will update in a new push, because that line has other changes in this review.

downloads/install.sh
```diff
@@ -112,6 +112,29 @@ fi
...
+
+# Create a variable that can be called on to question if an exit 1 is required (due to items already existing)
+exit_due_to_existing_item=0
+
+# Iterate through the list of items
+for item in "${items[@]}"; do
+    if [ -d "${current_dir}${item}" ]; then
+        # Update the variable because the found existing item now warrants an exit 1 after this loop
+        exit_due_to_existing_item=1
+        fancy_print 1 "Error: ${item} already exists, please delete ${current_dir}${item} and run the installer again."
```
tpmccallum, line 129:
```suggestion
        fancy_print 1 "Error: .${item} already exists, please delete ${current_dir}${item} and run the installer again."
```

itowlson (approved):
LGTM though would be great to fit in @karthik2804's change as well!

itowlson (approved):
:shipit:

tpmccallum:
New logic, check only for `spin` directory.
Screenshot 2023-09-19 at 13 46 41

Issue #893, opened 2023-09-19
`spin build` page should not recommend `RUST_LOG`


itowlson:
The bit on running `spin build` shows it with `RUST_LOG`.  This adds clutter; if what it shows is important to the user experience then it should be part of the user experience not an INFO log.

![image](https://github.com/fermyon/developer/assets/865538/1b02f42b-0f0f-43db-b04a-2d763dfcfb20)

PR #896, opened 2023-09-19
Update build command output to not include RUST_LOG

tpmccallum:

### DIFF
--- a/content/spin/build.md
+++ b/content/spin/build.md
@@ -151,13 +151,13 @@ Once the build commands are set up, running `spin build` will execute, sequentia
 
 ```bash
 $ spin build
-Building component rust-hello with `cargo build --target wasm32-wasi --release`
+Building component hello with `cargo build --target wasm32-wasi --release`
     Updating crates.io index
     Updating git repository `https://github.com/fermyon/spin`
 
     //--snip--
 
-    Compiling test-application v0.1.0 (rust-hello)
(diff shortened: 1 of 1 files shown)


tpmccallum:
Fixes https://github.com/fermyon/developer/issues/893

content/spin/build.md
```diff
@@ -41,7 +42,7 @@ command = "npm run build"
 
 If you generated the component from a Fermyon-supplied template, the `component.build` section should be set up correctly for you.  You don't need to change or add anything.
 
-> Different components may be built from different languages, and so each component can have its own build command.  In addition, some components may be precompiled into Wasm modules, and don't need a build command at all.  If a component doesn't have a build command, `spin build` just skips it.
+> Different components may be built from different languages, and so each component can have its own build command.  In addition, some components may be precompiled into Wasm modules, and don't need a build command at all.  If a component doesn't have a build command, `spin build` just skips it:
```
itowlson, line 45: The full stop was correct here. This is a side note; it does not lead into the subsequent tab block.
```suggestion
> Different components may be built from different languages, and so each component can have its own build command.  In addition, some components may be precompiled into Wasm modules, and don't need a build command at all.  If a component doesn't have a build command, `spin build` just skips it.
```

itowlson (changes requested):
The tiniest thing I'm afraid, just need to revert one change.

itowlson (approved):

PR #901, opened 2023-09-21
Document inferencing option defaults

mikkelhegn:
Sourced from: https://github.com/fermyon/spin/blob/09da960ee590a36170608737948f22e0d6de35b4/crates/llm/src/lib.rs#L386


karthik2804 (approved):

content/spin/serverless-ai-api-guide.md
```diff
@@ -53,7 +53,7 @@ The set of operations is common across all supporting language SDKs:
...
...performed on a specific model.<br /> <br />The name of the model is the first parameter provided (i.e. `llama2-chat`, `codellama-instruct`, or other; passed in as a `string`).<br /><br /> The second parameter is a prompt; passed in as a `string`.<br /><br /> The third parameter is a mix of float and unsigned integers relating to inferencing parameters in this order: <br /><br />- `max-tokens` (unsigned 32 integer) Note: the backing implementation may return less tokens. <br /> Default is 100<br /><br /> - `repeat-penalty` (float 32) The amount the model should avoid repeating tokens. <br /> Default is 1.1<br /><br /> - `repeat-penalty-last-n-token-count` (unsigned 32 integer) The number of tokens the model should apply the repeat penalty to. <br /> Default is 64<br /><br /> - `temperature` (float 32) The randomness with which the next token is selected. <br /> Default is 0.8<br /><br /> - `top-k` (unsigned 32 integer) The number of possible next tokens the model will choose from. <br /> Default is 40<br /><br /> - `top-p` (float 32) The probability total of next tokens the model will choose from. <br /> Default is 0.9<br /><br /> The result from `infer_with_options` is a `string` |
```
itowlson, line 56: Looking at the code I would definitely not describe the options record as a "list" - for many programmers that implies a vector or array kind of thing, rather than a record of named values.

I'm also concerned this could be a significant pain to maintain - that's a lot of text and embedded HTML markup squooshed onto one line, easy to get things mixed up!  Instead, I would pull out a separate table called something like "Inference Options" to hold the record info (same content, but now Markdownable), and have the `params` entry and description _refer_ to that table.  E.g.

```suggestion
| `infer_with_options`  | model `string`,<br />prompt `string`,<br />params `record` | `string`  | The `infer_with_options` is performed on a specific model.<br /> <br />The first parameter is the name of the model (e.g. `llama2-chat`, `codellama-instruct`, or other), passed in as a `string`.<br /><br /> The second parameter is a prompt, passed in as a `string`.<br /><br /> The third parameter is a record of options to control how the inferencing is done - see the Inferencing Options table for details. |
```

itowlson, line 56: The `Returns` column already says that it returns a string - we can drop that last sentence.

itowlson (changes requested):

itowlson:
More generally we should go back and review how we present APIs.  The tabular structure was, I think, adopted in a bit of a hurry for Spin 1.0 (and I think the blame for that is on me), and was already showing its limitations a few APIs ago.  Not only does it run hard into the limitations of Markdown tables, it doesn't provide room for talking about the language projections against the API itself.  (This was partly an artifact of the hurry.  Presenting the abstract API and then notes on the language projections was quicker than writing out multiple similar SDK surfaces at reference quality.  But it does have the advantage of focusing on the language neutral nature of the APIs-as-WIT-interfaces.)  So I'd be keen to revisit the API pages generally in the light of our greater experience, which would provide a better framework for APIs like this.  That's out of scope for this PR of course!

mikkelhegn:
@itowlson - This is good feedback, and I agree. The APIs are also different between Rust and JS/TS, as in the latter there is only one function, with an optional option argument.

This PR only added the default values for the options, and I believe your suggestion removes that again, is that true?

If we don't want to add those defaults value here now, then I'll close this PR, and we can add an issue to fix based on your comments. Let me know what you think.

itowlson:
@mikkelhegn My suggestion does not merely remove the default values - it removes ALL details of the options! ...with the recommendation to move all that info, including the default values, to a separate table.

But yes, I somehow missed that the big honkin' text was already there and you were only adding the defaults.  Doh!  Given that we already have the big cell, I'm okay with adding the defaults to that big cell,   Apologies for the misreading.

itowlson (approved):

mikkelhegn:
No worries @itowlson :-)

PR #906, opened 2023-09-25
Add redeploy error to FAQ

tpmccallum:

### DIFF
--- a/content/cloud/faq.md
+++ b/content/cloud/faq.md
@@ -110,7 +110,6 @@ To learn more about what feature support looks like for various programming lang
   - Fermyon Cloud only supports Fermyon DNS at this time for custom domains. To file a feedback request, please visit [github.com/fermyon/feedback](https://github.com/fermyon/feedback).
 
 - **Unable to redeploy an app on Fermyon Cloud?**
-<!-- @searchTerm “Error: No channel with app_id" -->
   - If `spin deploy` fails with an error (but `spin build —up` works) please delete the app from your [Fermyon Cloud](https://cloud.fermyon.com/) dashboard and then try `spin deploy` again.
 
 ## Next Steps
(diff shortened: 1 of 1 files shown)


content/cloud/faq.md
```diff
@@ -109,6 +109,9 @@ To learn more about what feature support looks like for various programming lang
 - **Can I configure external DNS for my custom domain on Fermyon Cloud?**
   - Fermyon Cloud only supports Fermyon DNS at this time for custom domains. To file a feedback request, please visit [github.com/fermyon/feedback](https://github.com/fermyon/feedback).
 
+- **Unable to redeploy an app on Fermyon Cloud?**
+  - If `spin build --up` (redeploying locally) works but `spin deploy` (redeploying an app to Fermyon Cloud) results in an error (i.e. `Error: No channel with app_id xxxx and name spin-deploy`). Please delete the app from your [Fermyon Cloud(https://cloud.fermyon.com/) dashboard and then try `spin deploy` again.
```
itowlson, line 113: suggestion to make it more specific that it's _this_ error message we're talking about

```suggestion
  - If `spin build --up` (redeploying locally) works but `spin deploy` (redeploying an app to Fermyon Cloud) results in the error `No channel with app_id xxxx and name spin-deploy`, please delete the app from your [Fermyon Cloud](https://cloud.fermyon.com/) dashboard and then try `spin deploy` again.
```

itowlson: Maybe even flip this around to "If `spin deploy` fails with the error ..., but `spin build --up` works, then..." - as someone scanning it is more likely to be looking for the "spin deploy fails" bit than the "spin up works" one

tpmccallum: I have removed the actual error message (because that is a bit waffly for a bullet point). But given that users search based on their errors I have added part of the error as a search term so a user’s search will plop them straight here.

itowlson: Okay something is not right here because that wasn't my suggestion... I think GitHub has messed up.  I was suggesting something like 

> If `spin build --up` (running locally) works but `spin deploy` (redeploying an app to Fermyon Cloud) results in the error `No channel with app_id xxxx and name spin-deploy`, please delete the app from your [Fermyon Cloud](https://cloud.fermyon.com/) dashboard and then try `spin deploy` again.

and then later suggested flipping the format.  I think we need to keep the error text.

itowlson (approved):

PR #913, opened 2023-10-02
add llm remote_http option

karthik2804:
Should we also add something about an example implementation being the `spin-cloud-gpu` plugin?

### DIFF
--- a/content/spin/dynamic-configuration.md
+++ b/content/spin/dynamic-configuration.md
@@ -241,6 +241,6 @@ url = "http://example.com"
 auth_token = "<auth_token>"
 ```
 
-Currently, using this runtime config to use remote compute requires an user to deploy their own LLM proxy service. A reference implementation for it can be found in the [`spin-cloud-gpu plugin repository`](https://github.com/fermyon/spin-cloud-gpu/blob/main/fermyon-cloud-gpu/src/index.ts). 
+Currently, the remote compute option requires an user to deploy their own LLM proxy service. Fermyon Cloud users can do this using the [`cloud-gpu` plugin](https://github.com/fermyon/spin-cloud-gpu).  If you prefer to create and deploy your own proxy service, you can find a reference implementation of the proxy protocol in the [`spin-cloud-gpu plugin repository`](https://github.com/fermyon/spin-cloud-gpu/blob/main/fermyon-cloud-gpu/src/index.ts). 
 
 By default, componenets will not have access to the LLM models unless granted explicit access through the `component.ai_models` entry in the component manifest within `spin.toml`. See [Serverless AI](/spin/serverless-ai-api-guide) for more details.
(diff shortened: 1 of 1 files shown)


itowlson (changes requested):
Looks good - just a few suggestions / omissions.  I agree you definitely need to mention / link the `cloud-gpu` plugin, otherwise it's all a bit theoretical!  I'm a bit torn between your current neutral approach (which is correct) versus leading off with cloud-gpu and including instructions (which is opinionated but in practice it's what this means today, and is more likely to jump out as an actual solution).  Although there are better places to do that - this page is pretty out-of-the-way!  What you have is fine, I am just thinking out loud.

I wonder if you want to add anything to the LLM tutorials or API guide drawing readers' attention to cloud-gpu?

content/spin/dynamic-configuration.md
```diff
@@ -227,3 +227,18 @@ Spin does _not_ create libSQL databases.  Use your hosting service's tools to cr
 The `default` database will still be defined, even if you add other databases.
 
 By default, components will not have access to any of these databases (even the default one). You must grant each component access to the databases that it needs to use. To do this, use the `component.sqlite_databases` entry in the component manifest within `spin.toml`. See [SQLite Database](/spin/sqlite-api-guide.md) for more details. 
+
+### LLM Runtime Configuration
+
+Spin provides an Large Language Model interface for interacting with LLMs for inferencing and embedding. The default host implementation is to use local CPU/GPU compute.  However, the Spin runtime configuration file (runtime-config.toml) can be updated to enable spin to use  remote compute using HTTP requests.
```
radu-matei, line 233: "Spin provides *a* Large Language Model interface"

content/spin/dynamic-configuration.md
```diff
@@ -227,3 +227,20 @@ Spin does _not_ create libSQL databases.  Use your hosting service's tools to cr
...
+The following is an example of how an application's `runtime-config.toml` file can be configured to use the remote compute option. Note the `type`, `url` and `auth_token` are set to `remote_http`, URL of the server and the auth token for the server. 
+
+```toml
+[llm_compute]
+type = "remote_http"
+url = "http://example.com"
+auth_token = "<auth_token>"
+```
+
+Currently, using this runtime config to use remote compute requires an user to deploy their own LLM proxy service. A reference implementation for it can be found in the [`spin-cloud-gpu plugin repository`](https://github.com/fermyon/spin-cloud-gpu/blob/main/fermyon-cloud-gpu/src/index.ts).
```
itowlson, line 244: wording got a bit intricate for this bear of very little brain
```suggestion
Currently, the remote compute option requires an user to deploy their own LLM proxy service. Fermyon Cloud users can do this using the [`cloud-gpu` plugin](https://github.com/fermyon/spin-cloud-gpu).  If you prefer to create and deploy your own proxy service, you can find a reference implementation of the proxy protocol in the [`spin-cloud-gpu plugin repository`](https://github.com/fermyon/spin-cloud-gpu/blob/main/fermyon-cloud-gpu/src/index.ts). 
```

itowlson (approved):
I have suggested some alternative wording for the LLM proxy paragraph but it's not a blocker

PR #937, opened 2023-10-18
Manifest v2 reference

itowlson:
This is just the reference.  We'll need to update other pages too but I wanted to get this out there.

@lann This is written from a single-trigger-type POV because that's what we have today - I know the format is more flexible but don't want to imply things that won't yet work.  Hopefully the rest of it is true though!

@tpmccallum A couple of ponderables for you (on top of wanting your feedback on the content!):

* I'm wondering if it would be good to have a "note for people familiar with Spin 1.x" style, distinct from our normal aside style, perhaps with an icon or something.  That's not intended to replace "if you're familiar with..." wording, but to draw the eye for folks who are migrating, and allow the eye to slide over for folks who are not!
* For this page specifically, I've put the v2 reference and v1 reference both on the one page.  The v1 reference is still relevant enough that I don't want to throw it out, but it does take up a lot of space and make the page structure harder to pick out at a glance.  (Yes, even harder than it was.)  Should we split this into two pages?  How would we want to present that in the TOC - same level in the Reference section, or a new "Spinning It Old Skool" section, or something?

✅ content/spin/v2/manifest-reference.md
✅ content/spin/v2/manifest-reference-v1.md

Screenshot 2023-10-30 at 07 49 57
Screenshot 2023-10-30 at 07 49 45

Screenshot 2023-10-30 at 07 52 06

### DIFF
--- a/content/spin/manifest-reference.md
+++ b/content/spin/manifest-reference.md
@@ -7,22 +7,17 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/manifest-refe
 ---
 - [Format](#format)
 - [Manifest Fields](#manifest-fields)
-- [The `application` Table](#the-application-table)
-- [The `application.trigger` Table](#the-applicationtrigger-table)
-  - [The `application.trigger.http` Table](#the-applicationtriggerhttp-table)
-  - [The `application.trigger.redis` Table](#the-applicationtriggerredis-table)
-- [The `variables` Table](#the-variables-table)
 - [The `trigger` Table](#the-trigger-table)
-  - [Common Fields for All `trigger.(type)` Tables](#common-fields-for-all-triggertype-tables)
-  - [Additional Fields for `trigger.http` Tables](#additional-fields-for-triggerhttp-tables)
-  - [Additional Fields for `trigger.redis` Tables](#additional-fields-for-triggerredis-tables)
-- [The `component` Table](#the-component-table)
-- [The `component.(id).build` Table](#the-componentidbuild-table)
+  - [The `trigger` Table for HTTP Applications](#the-trigger-table-for-http-applications)
+  - [The `trigger` Table for Redis Applications](#the-trigger-table-for-redis-applications)
+- [The `variables` Table](#the-variables-table)
+- [The `component` Tables](#the-component-tables)
+  - [The `component.trigger` Table for HTTP Applications](#the-componenttrigger-table-for-http-applications)
+  - [The `component.trigger` Table for Redis Applications](#the-componenttrigger-table-for-redis-applications)
+  - [The `component.build` Table](#the-componentbuild-table)
 - [Next Steps](#next-steps)
 
 This page describes the contents of the Spin manifest file, typically called `spin.toml`.
-
-> There are two versions of the manifest format. The manifest format described here (version 2) is recommended if you're using Spin 2.0 and above. The [previous format (version 1)](manifest-reference-v1.md) is supported on Spin 2.x for backward compatibility, and is the only format supported by Spin 1.x.
 
 ## Format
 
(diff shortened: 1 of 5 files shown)


tpmccallum:
Hi @itowlson 
Thanks, I am leaning into having V1 and V2 pages. This is a lot of content for one page. Also being mindful of folks on smaller devices and phones, trying to read this page :grimacing: 

Would you consider the separate page option? The 2-page option will allow us to remove the 30-odd instances of the words "Version 1" from that page (and its ToC). 
We could then create a separate menu item under the top-level `Reference` heading (and name the page and title to match)
- `Spin V1 Application Manifest`
- `Spin V2 Application Manifest`
A massive thanks for editing this page. A big undertaking that is greatly appreciated.
I am happy with 2-page approach.

itowlson:
I would _totally_ consider the separate page option!  Agree that v1 and v2 manifest side by side in the TOC are easiest for now.  I'll get that under way.

itowlson:
@tpmccallum Split done

content/spin/manifest-reference.md
```diff
@@ -27,42 +32,47 @@ The manifest is a TOML file, and follows standard TOML syntax.  See the [TOML do
...
+
+> If you're familiar with manifest version 1, note that the way trigger parameters map to components - for example, which component handles a particular HTTP route - is now defined on the _trigger_, not on the component. In the version 2 manifest, a `component` section specifies _only_ the Wasm module and the resources it needs.
+
+## The `application` Table
+
+| Name                    | Required?  | Type        | Value    | Example   |
+|-------------------------|------------|-------------|----------|-----------|
 | `name`                  | Required   | String      | The name of the application. This may be any string of alphanumeric characters, hyphens and underscores. | `"hello-world"` |
 | `version`               | Required   | String      | The version of the application. The must be a string of the form `major.minor.patch`, where each element is a number. | `"1.0.5"` |
```
lann, line 48: This is no longer mandatory for `spin up` at least.

```suggestion
| `version`               | Optional   | String      | The version of the application. The must be a string of the form `major.minor.patch`, where each element is a number. | `"1.0.5"` |
```

content/spin/manifest-reference.md
```diff
@@ -27,42 +32,47 @@ The manifest is a TOML file, and follows standard TOML syntax.  See the [TOML do
...
+
+> If you're familiar with manifest version 1, note that this is now optional. The trigger type to which the application responds is determined by the top-level `trigger` mappings; you only need to provide an `application.trigger` if you want to override the default global settings for the trigger type.
 
-### The `trigger` Table for HTTP Applications
+### The `application.trigger.http` Table
 
 | Name                    | Required?  | Type        | Value    | Example   |
 |-------------------------|------------|-------------|----------|-----------|
-| `type`                  | Required   | String      | Always `"http"` for HTTP applications. | `"http"` |
 | `base`                  | Required   | String      | The base path of the application. All component routes are relative to this. It allows multiple applications to be mounted under the same host. | `"/"` |
```
lann, line 65:
```suggestion
| `base`                  | Optional   | String      | The base path of the application. All component routes are relative to this. It allows multiple applications to be mounted under the same host. | `"/"` |
```

content/spin/manifest-reference.md
```diff
@@ -119,13 +121,38 @@ Each table in the `component` array contains the following fields:
...
+### Additional Fields for `trigger.redis` Tables
 
 | Name                    | Required?  | Type        | Value    | Example   |
 |-------------------------|------------|-------------|----------|-----------|
 | `channel`               | Required   | String      | The Redis channel which this component handles. Messages on this channel will cause the component to execute. | `"purchases"` |
 
-### The `component.build` Table
+## The `component` Table
+
+The keys of the `component` table are user-defined, and are usually written as part of the table syntax e.g. `[component.my-component]`.  The value of each key is another table with the fields shown in the following table.
```
lann, line 132: > The keys of the `component` table are user-defined,

This threw me for a loop for a minute thinking that it was referring to the inner table keys and would probably be even more confusing to anyone less familiar with TOML terminology. Perhaps clarify something like:

"The keys of the `component` table (e.g. `my-component` in `[component.my-component]`)"

content/spin/manifest-reference.md
itowlson: Oops!  Thanks

content/spin/manifest-reference.md
itowlson: Good catch - thank you.  I've updated as you suggest, and the resulting text is:

> The keys of the `component` table - such as the `my-component` in `[component.my-component]` - are user-defined, and are usually written as part of the table syntax e.g. `[component.my-component]`.

It's clear, but do you think it's okay that it repeats the "[component.my-component]" thing?  I can play around with the wording if not.

lann: More clear but yeah a bit awkward.

Also I just noticed:

> table with the fields shown in the following table.

which might cause table-madness. Maybe "table with the following fields:"

itowlson: Huh, that is indeed a lot of tables.  Thanks!

tpmccallum:
Thinking a bit more about this PR, what are your thoughts around keeping both the V1 and V2 pages alive on the `latest` equivalent? I think having both pages present and in the menu (on the frozen content that will be copied off into the path of `/v1` and also on the current docs) will give us a chance to explain that there is a `v1` and a `v2` and they are a bit different (and the application manifest is how you configure an application etc. etc.). Long story short, this PR can go ahead as planned and remain a part of the current content moving forward. Does that sound good?

itowlson:
Yes, this is one of the cases where I believe we should (at least initially) retain v1 content on latest.  The v1 manifest works fine with v2, and users may be maintaining and updating v1 applications.

tpmccallum:
So essentially we would:
- merge this PR,
- then copy off the frozen 1.0 flavoured content,
- then update current content to be 2.0 flavoured.

itowlson:
I am a bit confused now.  We'd probably want to snapshot the v1 content _before_ merging this PR, wouldn't we?

(Also we don't want to merge this PR until we release v2.  Unless we have the v2 content merged, but only surface v1 until we are ready to swap v2 in as latest?  Maybe we need to chat.  MAYBE WE NEED A WHITEBOARD.)

itowlson:
This is going to need some sidebar surgery too.  I backed the TOC change out from here because the v1 reference link would have been bad, but it will need to be reinstated once we figure out the strategy.

itowlson:
bump

tpmccallum (approved):
LGTM
I will make a note of the sidebar surgery required as part of the flip-switch procedure.

PR #949, opened 2023-10-24
Create v2 dir and update dates in frontmatter

tpmccallum:
Preliminary PR to create a `v2` folder and change dates in all markdown files.

Depending on official release, we might have to re-tweak the dates. You can use the following script to do so:

```bash
#!/bin/bash
for file in developer/content/spin/v2/*.md; do
    if [ -f "$file" ]; then
        sed -i '/^date/s/.*/date = "2023-11-02T16:00:00Z"/' $file
    fi
done
```


itowlson (approved):
Update links in later pass :shipit:

PR #953, opened 2023-10-25
Non breaking path changes in preparation for v2 folder structure

tpmccallum:
These non-breaking changes support the move to having a separate `v1` folder and a separate `v2` folder when Spin 2.0 is announced. 

This PR corrects the misuse of absolute URLs to developer documentation pages and removes relative links to the redundant `common` directory (which is now handled dynamically by Spin redirects). It also uses the `./` in spin pages to access other pages that are also in the spin folder (i.e. uses ./pagename instead of the original /spin/pagename).
Updates only the v2 folder to include relative links to the v2 GitHub source in the frontmatter.

These are designed to be merged now with no consequences to the current operation of the docs but pave the way for the following changes that are out of scope for this PR:
- changes to `spin_sidebar.hbs` (dropdown for v1 and v2 options) 
- modification of redirects in the `spin.toml` file.

Linting files
- [x] All files pass
image

Checking URL endpoint for broken links
- [x] URLs
image


itowlson:
It's a bit hidden because there was already one link broken on main, but this has about 10 broken links - might be worth looking into whether they are legit.

tpmccallum:
Thanks @itowlson 
I will take a look now (on localhost)

itowlson (approved):
I didn't close-read the second half of v2 but this looks good as far as I can tell!  Thanks for battling through it!

itowlson:
Oh, yay, the Vault config link is because it moved to https://github.com/fermyon/spin/tree/main/tests/http/vault-variables-test as part of v2 renaming.  But for v1 we probably want to keep the link pointing to the old name e.g. https://github.com/fermyon/spin/tree/v1.5/tests/http/vault-config-test

tpmccallum:
Thanks @itowlson 
- I kept v1 to config test and only updated the v2 (to the new vault URL in GH)
- fixed the broken links (were caused by Bartholomew's page that dynamically loads "Contributing" via the `spin` folder, seeing the wrong paths)
- Also fixed some other syntax issues and all tests out 100% (as shown in description above).

itowlson (approved):
Thanks for figuring those out!  LGTM

PR #960, opened 2023-10-26
Adding serverless ai hello world to v2 folder

tpmccallum:
This will need some sidebar surgery also @karthik2804 
This new page was added to the standard docs after the `cp spin/ v2` was performed.
Will there be a separate `spin_sidebar.hbs` file for `v2`?


tpmccallum:
@karthik2804 ’s PR already added this page so this can be close.

PR #975, opened 2023-10-27
chore: Update all TinyGo SDK examples to v2

adamreese:


adamreese:
Links will be broken until Spin has been released

tpmccallum:
Hi @adamreese 
Thanks so much for these updates.

`v2`
The `v2` files will not see the light of day until the date in the frontmatter arrives (which at this stage is 4 pm 2nd Nov. UTC).
This can be merged safely in anticipation of the links resolving then (because Spin v2 will be released).

`v1`
Regarding v1 files, they are considered legacy and will be frozen off into the `v1` directory to provide ongoing support for v1 (should anyone choose to remain on v1). For this reason the contents in `v1` needs to stay compatible with existing v1 apps (so you will notice that I have made this specific change to your PR).

The v1 pages will have a canonical reference tag set (to point to the new v2 page equivalent) making v2 the canonical page and v1 just for reference (should anyone want to deliberately visit those outdated pages).

Hope this makes sense and thanks again, great appreciate this work!

tpmccallum (approved):
This great @adamreese, thank you!

PR #991, opened 2023-10-31
Update `allowed_outbound_hosts` to final behaviour

itowlson:
Outbound restrictions for databases and Redis are applied on the basis of SDK version not manifest version.  Therefore, there is no need to update v1 SDK docs to refer to outbound allow-lists.  (There is some nuance about what is a v1 SDK vs what is a v2 SDK but that's best avoided.  For now, we want people to err on the side of allow-listing their hosts.)

As v2 SDKs may be used with v1 manifests, we are adding `allowed_outbound_hosts` to the v1 manifest (but only in Spin 2).


content/spin/v2/manifest-reference.md
```diff
@@ -144,7 +144,7 @@ The value of each key is a table with the following fields.
...
-| `allowed_outbound_hosts` | Optional  | Array of strings | The addresses to which the Wasm component is allowed to send network requests. Each entry must contain both a name (or IP address) and a port in `name:port` format. For Redis, MySQL, and PostgreSQL, the component may only contact the addresses listed here; for HTTP, the component may also contact hosts in `allowed_http_hosts`. | `["example.com:4567", "127.0.0.1:5000"]` |
+| `allowed_outbound_hosts` | Optional  | Array of strings | The addresses to which the Wasm component is allowed to send network requests. Each entry must contain both a name (or IP address) and a port in `name:port` format. For Redis, MySQL, and PostgreSQL, the component may only contact the addresses listed here; for HTTP, the component may also contact hosts in `allowed_http_hosts`. If this field is omitted, Spin 1 components default to allow-all access for backward compatibility; otherwise, no outbound access is permitted except for HTTP. | `["example.com:4567", "127.0.0.1:5000"]` |
```
rylev, line 147:
```suggestion
| `allowed_outbound_hosts` | Optional  | Array of strings | The addresses to which the Wasm component is allowed to send network requests. Each entry must contain a scheme (one of either "http", "https", "redis", "mysql", or "postgres"), a host name (or IP address) and an optional port or range of ports in `scheme://name:port` format. The component may only contact the addresses listed here. If the port is omitted, the default port for the scheme is defaulted to (e.g., 'https://' defaults to port 443). The special host "self" may be used to allow calls back to the current application. If this field is omitted, Spin 1 components default to allow-all access for backward compatibility. | `["redis://example.com", "https://127.0.0.1:5000"]` |
```

content/spin/v2/manifest-reference.md
```diff
@@ -144,7 +144,7 @@ The value of each key is a table with the following fields.
 |                         |            | Table       | * A file or directory to be made available to the Wasm module at a specific path. This must be a table containing a `source` field for the file or directory relative to the manifest file, and a `destination` field containing the absolute path at which to make it available. | `{ source = "assets/images", destination = "/pictures" }` |
 | `exclude_files`         | Optional   | Array of strings | Any files or glob patterns that should _not_ be available to the Wasm module at runtime, even though they match a `files` entry. | `[assets/images/test/**/*.*]` |
 | `allowed_http_hosts`    | Optional   | Array of strings | The host names or addresses to which the Wasm component is allowed to send HTTP requests. If the name includes a port, the Wasm module can send requests only to that port; otherwise, the Wasm module can send requests only to the default port for the scheme it uses. The special string `insecure:allow-all` permits the module to send HTTP requests to _any_ host, but is intended for development use only; some deployment environments may decline to honour it. | `["example.com", "localhost:8081"]` |
```
rylev, line 146:
```suggestion
| `allowed_http_hosts`    | Deprecated Optional   | Array of strings | Deprecated in favor of `allowed_outbound_hosts`. The host names or addresses to which the Wasm component is allowed to send HTTP requests. If the name includes a port, the Wasm module can send requests only to that port; otherwise, the Wasm module can send requests only to the default port for the scheme it uses. The special string `insecure:allow-all` permits the module to send HTTP requests to _any_ host, but is intended for development use only; some deployment environments may decline to honour it. | `["example.com", "localhost:8081"]` |
```

I very much think we should mark this as deprecated and perhaps move it to the end of the list. I think it would be unfortunate if new users find this documentation and mistakenly think that this is the right option to use.

content/spin/v2/manifest-reference-v1.md
```diff
@@ -106,6 +106,7 @@ Each table in the `component` array contains the following fields:
...
 | `allowed_http_hosts`    | Optional   | Array of strings | The host names or addresses to which the Wasm module is allowed to send HTTP requests. If the name includes a port, the Wasm module can send requests only to that port; otherwise, the Wasm module can send requests only to the default port for the scheme it uses. The special string `insecure:allow-all` permits the module to send HTTP requests to _any_ host, but is intended for development use only; some deployment environments may decline to honour it. | `["example.com", "localhost:8081"]` |
+| `allowed_outbound_hosts` | Optional  | Array of strings | The addresses to which the Wasm component is allowed to send network requests. Each entry must contain both a name (or IP address) and a port in `name:port` format. For Redis, MySQL, and PostgreSQL, the component may only contact the addresses listed here; for HTTP, the component may also contact hosts in `allowed_http_hosts`. If this field is omitted, Spin 1 components default to allow-all access for backward compatibility; otherwise, no outbound access is permitted except for HTTP. | `["example.com:4567", "127.0.0.1:5000"]` |
```
rylev, line 109: > Otherwise, no outbound access is permitted except for HTTP.

I found this wording confusing. If the `allowed_outbound_hosts` is set, traffic is allowed to anything that the value configured permits. The way it's currently worded sounds like all traffic will be blocked no matter what. Also, http traffic is not openly permitted, it's also blocked unless allowed by `allowed_outbound_hosts` or `allowed_http_hosts`.

itowlson:
I've reworked this significantly based on the changes in the latest Spin PRs.  Would be good to nail it ASAP @rylev if you are still around.

content/spin/v2/http-outbound.md
```diff
@@ -149,27 +149,27 @@ You can find a complete example for using outbound HTTP in the [Spin repository
...
 > This doesn't work in Redis components because Spin uses the incoming HTTP request to determine the current host.
 
-You must still grant permission by including `self` in `allowed_http_hosts`:
+You must still grant permission by including `self` in `allowed_outbound_hosts`:
 
 ```toml
-allowed_http_hosts = ["self"]
-```
\ No newline at end of file
+allowed_outbound_hosts = ["self"]
```
rylev, line 174:
```suggestion
allowed_outbound_hosts = ["https://self"]
```

rylev (changes requested):
Just one small correction!

content/spin/v2/http-outbound.md
itowlson: `self` is intended to be portable across localhost and Cloud (or other host).  local is usually http not https.  Should this therefore be `["http://self", "https://self"]`?

(At one point I saw something that special cased `self` because it _only_ works with HTTP, but clearly that didn't survive!)

itowlson: @rylev done, thanks

itowlson:
@tpmccallum Unless @rylev is around, I think you might need to override the request changes (if it looks okay to you) - I'm pretty sure I've done what Ryan needs but it's late where he is but I don't want to wait a full 24-hour cycle before merging this.  We can of course make further changes, if he has them, in a follow-up PR.

rylev (approved):

PR #993, opened 2023-11-01
All Structural Changes (and some content) For Spin 2 Documentation Release - approved and ready

karthik2804:
I have created a new commit to your v2 update that adds a canonical URL entry into the [extra] section of every v1 .md file that has a canonical v2 equivalent.
https://github.com/fermyon/developer/pull/993/commits/8e3cd25baaff7678b8ec878cf7578639029375f0

### DIFF
--- a/config/site.toml
+++ b/config/site.toml
@@ -2,6 +2,7 @@ title = "Fermyon Developer"
 base_url = ""
 about = "Fermyon Developer — the home for you as a developer using Fermyon projects and products."
 index_site_pages = ["sitemap", "home", "changelog", "hub_list_api"]
+prepend_route_info = true
 
 [extra]
 copyright = "Fermyon"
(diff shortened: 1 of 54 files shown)


tpmccallum:
Thanks @karthik2804 will test this out now and comment below.

tpmccallum:
Hi @karthik2804 
The canonical HTML elements are not in the files in the `/spin/v1/` path.
Do you need more information about how to create these or how they should look/work?
Basically, every page in the `v1` path (that has a similar `v2` page; which is most of them at this point of divergence) should have the HTML element that points to the equivalent `v2` page. This is a really important part of the SEO once we flip the switch to 2.0.
Other than the canonical work this is looking really great. Testing well on localhost.

tpmccallum (approved):
LGTM

PR #1010, opened 2023-11-03
initial draft -adding embeddable spin up hub cards

karthik2804:
To add cards - use the following shortcode format 

```
{{suh_cards}}
{{card_element <type(sample|template)> <title> <summary> <URL> <tags>}}
{{blockEnd}}
```

All the arguments to `card_element` should be inside double quotes. The tags should be separated by a comma. 

Note: This requires enabling shortcodes by adding `enable_shortcodes = true` to the frontmatter.


flynnduism:
@karthik2804 these look great - and the way the snippets are configed is 👏🏻

flynnduism (approved):

itowlson:
@karthik2804 Could you add the "how to embed" info (ideally with an example) to the read-me or contributing doc or wherever it seems to fit please?  Thanks!

(future thought: what if there was a shortcode cheat sheet with _all_ the codes in it)

PR #1046, opened 2023-11-16
Add TS example of handling POST data from a router.

technosophos:


technosophos:
Huge thanks to @bacongobbler for finding a couple of potential fixes.

itowlson (approved):

karthik2804:
@technosophos this looks great. Out of curiosity, would it be better if we add an example of using the router with the `spin-component-route` along with it and make it a more general example of the usage of the router?

technosophos:
@karthik2804 Yeah, that is a good idea. Let me add that.

technosophos:
Okay, @karthik2804 -- updated in both

PR #1055, opened 2023-11-26
fix sitemap to point at preview pages instead of api

karthik2804:


flynnduism:
LGTM, but would wait to see what @tpmccallum thinks before shipping

tpmccallum:
Thanks @karthik2804 
Good spotting.
I wondered if it would be better to redirect at the `spin.toml` level so that the sitemap naturally renders the correct URLs.
Is it technically possible to implement a `path_prefix` redirect in the manifest?

tpmccallum (approved):
Thanks so much @karthik2804

tpmccallum:
Just double-checked it locally. Looks great!
Thanks again @karthik2804 
Screenshot 2023-11-27 at 10 54 40

PR #1056, opened 2023-11-26
update klaro config

flynnduism:
Fixing an issue with Klaro js cookie banner, ensuring analytics don't load


static/js/main.js
karthik2804, line 1: This is a generated file  - the changes need to go to `static/js/src/main.js` and then to build `npm run bundle-scripts`

karthik2804 (approved):

karthik2804 (approved):

PR #1062, opened 2023-11-28
Fix manifest v2 component variables docs bug

rylev:
Fixes a few issues:
* The example for the component specific variables configuration was incorrect and has been fixed
* The wording around the component table was confusing so I tried to improve it


content/spin/v2/manifest-reference.md
```diff
@@ -147,8 +147,8 @@ The value of each key is a table with the following fields.
...
-| `variables`             | Optional   | Table       | Dynamic configuration values to be made available to this component. The table keys are user-defined; the values must be strings, and may use template notation as described under [Dynamic Configuration](dynamic-configuration). | `[component.variables]`<br />`api_base_url = "https://{{ api_host }}/v1"` |
+| `build`                 | Optional   | Table       | The command that `spin build` uses to build this component. See [The `component.(id).build` Table](#the-componentidbuild-table) below. | `[component.my-component.build]`<br />`command = "npm run build"` |
+| `variables`             | Optional   | Table       | Dynamic configuration values to be made available to this component. The table keys are user-defined; the values must be strings, and may use template notation as described under [Dynamic Configuration](dynamic-configuration). | `[component.my-component.variables]`<br />`api_base_url = "https://{{ api_host }}/v1"` |
```
itowlson, line 151:
```suggestion
| `variables`             | Optional   | Table       | Dynamic configuration values to be made available to this component. The table keys are user-defined; the values must be strings, and may use template notation as described under [Dynamic Configuration](dynamic-configuration). | `[component.cart.variables]`<br />`api_base_url = "https://{{ api_host }}/v1"` |
```

content/spin/v2/manifest-reference.md
```diff
@@ -129,7 +129,7 @@ Each array entry contains a mix of common fields and trigger-specific fields.
 
 ## The `component` Table
 
-The keys of the `component` table, usually written as part of the table syntax e.g. `[component.my-component]`, are user-defined.  (In the preceding example, the key is `my-component`). Component names must be kebab-cased, i.e. the only permitted separator is a hyphen.
+The keys of the `component` table, usually written using TOML table syntax e.g. `[component.my-component]`, are ids of components.  Component ids must be kebab-cased, i.e. the only permitted separator is a hyphen.  In the following example, `my-component` is used as an example component id.
```
itowlson, line 132: I find "using TOML table syntax" less specific than "as part of the table syntax", because the latter guides me more to it being a _part_ of something rather than the usual key-value table syntax.  But neither is ideal.  We can and should wordsmith this, but perhaps it would be better to decouple that from the "missing component name in the variables example" fix as I can see a _lot_ of tweaking being needed to get this right, whereas the fix we can just knock off.

itowlson, line 132: We should probably come to an agreement on IDs vs names and apply it throughout.  I know I have gone back and forth.  In v1 I think it was mostly 'ID', but I believe v2 favoured 'name' because of alignment with component model terminology?  I am not sure.

(Also 'name' finesses needing to agree a standard on 'id' vs 'ID' _grin_)

content/spin/v2/manifest-reference.md
```diff
@@ -147,8 +147,8 @@ The value of each key is a table with the following fields.
...
 | `environment`           | Optional   | Table       | Environment variables to be set for the Wasm module. This is a table. The table keys are user-defined; the values must be strings. | `{ DB_URL = "mysql://spin:spin@localhost/dev" }` |
-| `build`                 | Optional   | Table       | The command that `spin build` uses to build this component. See [The `component.(id).build` Table](#the-componentidbuild-table) below. | `[component.cart.build]`<br />`command = "npm run build"` |
-| `variables`             | Optional   | Table       | Dynamic configuration values to be made available to this component. The table keys are user-defined; the values must be strings, and may use template notation as described under [Dynamic Configuration](dynamic-configuration). | `[component.variables]`<br />`api_base_url = "https://{{ api_host }}/v1"` |
+| `build`                 | Optional   | Table       | The command that `spin build` uses to build this component. See [The `component.(id).build` Table](#the-componentidbuild-table) below. | `[component.my-component.build]`<br />`command = "npm run build"` |
```
itowlson, line 150: I prefer the short name in the table - horizontal space is at a premium here and the important bit is at the end of the line.

```suggestion
| `build`                 | Optional   | Table       | The command that `spin build` uses to build this component. See [The `component.(id).build` Table](#the-componentidbuild-table) below. | `[component.cart.build]`<br />`command = "npm run build"` |
```

tpmccallum:
These are fixed in https://github.com/fermyon/developer/pull/1064
Thanks @rylev and sorry for the overlap.

Issue #540, opened 2023-04-11
Update Python docs (i.e. list existing functions to get headers when working with request/response)


tpmccallum:
A question in Discord asked "is there documentation somewhere for the spin python sdk ... api docs", as apposed to a walkthrough/tutorial?
 For example, the Discord comment asks what's available in relation to things like working with request/response object i.e. is there an official way of getting headers (for example; as per the Discord message "... some other existing `.get_headers()` or `.to_dict` functionality I didn't discover?"

PR #1088, opened 2023-12-12
Update python-components.md

tpmccallum:

Closes https://github.com/fermyon/developer/issues/540


tpmccallum:
Just a quick mention about how to access response object when using HTTP-PY
Also love the fact that we can print to the console and do `dir(varname)` and `type(varname)` to helpful.

dicej (approved):

PR #1111, opened 2023-12-28
Fixed "qr code" problem in hub search

### LINKED ISSUE #1097: Hub search
Hub search misses obviously relevant cards.

For example, I searched for `qr code`, which I think is pretty specific and which I think has only one relevant card in the Hub.

That card was 15th in the search results.

image

(In case you think 'code' might have confused it, a search for `qr` had the exact same result except for "checklist app" and "silly walks" getting swapped around.  Go figure...)

Other examples are "news" missing the news summariser sample, and "dog feeding" missing the dog feeding sample.  We can tag our way out of any individual hole (although the risk is that then card displays get cluttered with large numbers of tags) but the request here is to think about a site-level search design that better understands titles and content.

VamshiReddy02:
Fixes #1097 

Result:
Screenshot 2023-12-28 at 2 12 58 PM

### DIFF
--- a/content/api/hub/sample_redirect.md
+++ b/content/api/hub/sample_redirect.md
@@ -14,7 +14,7 @@ last_updated = "2023-08-14T00:00:00Z"
 spin_version = ">v1.0"
 summary =  "A sample of using mulitple components to create an app"
 url = "https://github.com/mikkelhegn/redirect"
-keywords = "rust, static content, components, html"
+keywords = "rust, static content, components, html, qr code"
 
 ---
 


itowlson (approved):
I'm fine with the change, but #1097 was intended to raise a broader issue around Hub search, using "qr code" just as an example.  (For example, "news" and "dog feeding" both fail to find cards that are _right there in the top line._)  Better tagging, as you've done here, can mitigate it, but the real intent in #1097 was to ask for site-level improvements so search didn't miss these obvious matches.  I'll update #1097 to make this clearer.

PR #1116, opened 2024-01-03
Remove links to Rust target spec docs

itowlson:
Fixes #1114.

In this PR I've opted to remove the broken links altogether, because the moved pages seemed more scary than helpful to someone just wanting to install or use the target.  If other folks think the links did add value, I can restore them to the new location.

Content must go through a pre-merge checklist.

## Pre-Merge Content Checklist

This documentation has been checked to ensure that:

- [ ] The `title, `template`, and `date` are all set
- [ ] Does this PR have a new menu item (anywhere in `templates/*.hbs` files) that points to a document `.md` that is set to publish in the future? If so please only publish the `.md` and `.hbs` changes in real-time (otherwise there will be a menu item pointing to a `.md` file that does not exist)
- [ ] File does not use CRLF, but uses plain LF (hint: use `cat -ve <filename> | grep '^M' | wc -l` and expect 0 as a result) 
- [ ] Has passed [`bart check`](https://developer.fermyon.com/bartholomew/quickstart)
- [ ] Has been manually tested by running in Spin/Bartholomew (hint: use `PREVIEW_MODE=1` and run `npm run styles` to update styling)
- [ ] Headings are using Title Case
- [ ] Code blocks have the programming language set to properly highlight syntax and the proper copy directive
- [ ] Have tested with `npm run test` and resolved all errors
- [ ] Relates to an existing (potentially outdated) blog article? If so please add URL in blog to point to this content.

PR #1117, opened 2024-01-03
add empty href tag to fix macos chrome bug

### LINKED ISSUE #1107: Tag filter in Developer Hub removes itself in Chrome
When trying to filter by tags in Developer Hub, the page refreshes and removes the filter entirely. 

https://github.com/fermyon/developer/assets/28018084/841f4537-a7d8-4958-adaa-fa9f45feec50

karthik2804:
fixes #1107 

Weird bug in Chrome on macos where a `<a>` tag without a `href` was causing the page to reload. We should potentially not be using the `a` tag where it is not a link but that involves stylistic changes which can be done in a future PR. 

cc: @flynnduism


itowlson (approved):

itowlson:
CI errors are per #1116 and can be overridden

flynnduism (approved):

PR #1119, opened 2024-01-03
Trying to fix the overall performance of the hub search

VamshiReddy02:
fixes: #1097 

Hey @itowlson, can you review this PR? I took up your suggestion from my last PR i.e., #1111. Please let me know if any further changes need to be made.

### DIFF
--- a/spin-up-hub/src/components/ContentListing.vue
+++ b/spin-up-hub/src/components/ContentListing.vue
@@ -24,13 +24,13 @@ export default {
             if (this.searchTerm) {
                 let updatedQuery = this.searchTerm
                     .split(" ")
-                    .map(word => word + '^2 ' + word + '* ')
+                    .map(word => word + '^2 ' + word + '*')
                     .join(' ');
                 let result = this.searchIndex.search(updatedQuery)
                 //get only confident results
                 let matches = []
                 result.map(k => {
-                    if (k.score < 1) {
+                    if (k.score < 5) {
                         return
                     }
                     matches.push(data.find(docs => k.ref == docs.id))
--- a/spin-up-hub/src/components/ContentListing.vue
+++ b/spin-up-hub/src/components/ContentListing.vue
@@ -24,13 +24,13 @@ export default {
             if (this.searchTerm) {
                 let updatedQuery = this.searchTerm
                     .split(" ")
-                    .map(word => word + '^2 ' + word + '* ' + word + '~2')
+                    .map(word => word + '^2 ' + word + '* ')
                     .join(' ');
                 let result = this.searchIndex.search(updatedQuery)
                 //get only confident results
                 let matches = []
                 result.map(k => {
-                    if (k.score < 0.5) {
+                    if (k.score < 1) {
                         return
                     }
                     matches.push(data.find(docs => k.ref == docs.id))
--- a/spin-up-hub/src/store.js
+++ b/spin-up-hub/src/store.js
@@ -115,25 +115,22 @@ const store = createStore({
       })
     },
     async loadSearchData(context) {
-      let data = await fetch(import.meta.env.VITE_API_HOST + "/static/hub-index-data.json");
-      let documents = await data.json();
+      let data = await fetch(import.meta.env.VITE_API_HOST + "/static/hub-index-data.json")
+      let documents = await data.json()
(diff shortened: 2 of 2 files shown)


itowlson:
@VamshiReddy02 I'm not qualified to review the detailed behaviour, but it's definitely the kind of thing I was thinking of!  Thanks!  @karthik2804 is much more familiar with this side of things and can hopefully take a look.

spin-up-hub/src/store.js
```diff
@@ -115,22 +115,25 @@ const store = createStore({
       })
     },
     async loadSearchData(context) {
-      let data = await fetch(import.meta.env.VITE_API_HOST + "/static/hub-index-data.json")
-      let documents = await data.json()
+      let data = await fetch(import.meta.env.VITE_API_HOST + "/static/hub-index-data.json");
```
karthik2804, line 118: nit: I would not add semicolons here just to be consistent with the rest of the code.

karthik2804 (changes requested):
Thanks for this @VamshiReddy02. This is a great starting place to address the search issue.

 I did some investigating into the query generation side of things. Based on the investigation, a couple of extra things could be tweaked in addition to this PR to improve the performance.
 
1. Remove the fuzzy matching in the query or at least limit it to 1 character [here](https://github.com/fermyon/developer/blob/cafc63a0aca48029fcf58ed9e7b1e4458cdaad7c/spin-up-hub/src/components/ContentListing.vue#L27)

1. Increase the threshold while filtering the matches to something higher than the current `0.5` which is really low [here](https://github.com/fermyon/developer/blob/cafc63a0aca48029fcf58ed9e7b1e4458cdaad7c/spin-up-hub/src/components/ContentListing.vue#L33)

Would you be open to adding those changes to this PR?

VamshiReddy02:
Hey @karthik2804, thank you for the feedback. I made some changes which were requested. I'm still doubtful about the fuzzy matching and Increase in the threshold changes, can you check and let me know if there are any changes to be made?

spin-up-hub/src/components/ContentListing.vue
```diff
@@ -24,13 +24,13 @@ export default {
             if (this.searchTerm) {
                 let updatedQuery = this.searchTerm
                     .split(" ")
-                    .map(word => word + '^2 ' + word + '* ' + word + '~2')
+                    .map(word => word + '^2 ' + word + '* ')
```
karthik2804, line 27:
```suggestion
                    .map(word => word + '^2 ' + word + '*')
```

spin-up-hub/src/components/ContentListing.vue
```diff
@@ -24,13 +24,13 @@ export default {
...
                     .join(' ');
                 let result = this.searchIndex.search(updatedQuery)
                 //get only confident results
                 let matches = []
                 result.map(k => {
-                    if (k.score < 0.5) {
+                    if (k.score < 1) {
```
karthik2804, line 33: This is the right place but I would go higher with all the boosts we have, in my experiments something around 5 was good.

karthik2804 (commented):
Just a couple of small tweaks but looks great overall! Thanks once again for handling this!

VamshiReddy02:
@karthik2804 please review this PR.

spin-up-hub/src/store.js
```diff
@@ -118,15 +118,15 @@ const store = createStore({
...
-        this.field('title')
+        this.field('title', { boost: 100 })
         this.field('content')
         this.field('language', { boost: 10 })
-        this.field('author'),
-          this.field('tags', { boost: 100 })
-        this.field('keywords', { boost: 100 })
+        this.field('author', { boost: 10 }),
+        this.field('tags', { boost: 10 })
+        this.field('keywords', { boost: 10 })
```
karthik2804, line 126: I think we should leave these fields as is, we want to match tags and keywords - keywords are a way to make sure we can specify alternative terms from tags. If someone wants to search for a particular author, they can search something along the lines of `author:<name>`

VamshiReddy02: Got it

karthik2804 (approved):

karthik2804:
Thanks for this - this is great! I will merge once CI is done

PR #1120, opened 2024-01-03
replace blur with dim transition

### LINKED ISSUE #884: Mystery black border flash when dropping down Docs menu
I went to the docs homepage and moved the mouse over the Docs drop-down.  When I did this, the background gradually blurred, and a strange blurry black border started to appear around the edges of the (visible area of the) page.  After a moment the blur reached its maximum blurriness at which point the black border suddenly vanished.

MS Edge on Windows 10.

image

karthik2804:
Fixes #884 

The blur overlay seemed odd and at times slow, so I decided to replace it with a dim animation. 

image

If that seems acceptable, we can merge the PR else I will look at the blur animation.

### DIFF
--- a/static/sass/developer-include-topbar.scss
+++ b/static/sass/developer-include-topbar.scss
@@ -14,10 +14,10 @@
         margin-right: -2px;
         position: relative;
         border: none;
-        
+
         .logo {
-            margin-top: -4px !important; 
-            
+            margin-top: -4px !important;
+
             svg {
                 height: 30% !important;
                 min-height: auto !important ;
@@ -45,7 +45,7 @@
             right: 0;
         }
     }
-    
+
     .navbar-menu {
         padding-left: 1.33vw;
         padding-right: 1vw;
@@ -173,15 +173,15 @@
     // docs project dropdown
     .dropdown {
         position: relative;
-    
+
         :hover .dropdown-content {
             display: block;
         }
-    
+
         .logo-project {
             display: none;
         }
-    
+
         &:hover+.dropdown-backdrop {
             position: fixed;
             width: 100vw;
@@ -190,9 +190,9 @@
             left: 0;
             z-index: -1;
         }
-    
+
     }
-    
+
     .dropdown-content {
         display: inline-block;
         min-width: 200px;
@@ -200,7 +200,7 @@
         background: transparent;
         box-shadow: none;
         margin: 2rem 0 0 0.3335rem;
-    
+
         a {
             color: #213762;
             font-size: 1.125rem;
@@ -210,7 +210,7 @@
             font-weight: bold;
             margin-right: 1.3335rem;
             position: relative;
-    
+
             &.is-active {
                 &::after {
                     content: "";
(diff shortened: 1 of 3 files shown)


itowlson:
The current dim level is maybe a bit dramatic for my taste (e.g. the drop-down now feels isolated and unmoored).  Could we try it with a more gentle dimming?  The user's attention is already in the right part of the screen, we don't need to over-emphasise the focus assist.  Thanks though, the direction seems nice and this is very much a nit!

(by the way, the formatting changes made it a bit laborious to see what the _actual_ CSS change was - it would be good to do those in a separate commit or PR so that someone coming back to this can easily see exactly what change was made)

karthik2804:
We could reduce the intensity of the dim. I will go back and remove the formatting fixes on the PR (I am going to blame it on my auto linter).

static/sass/developer-include-topbar.scss
```diff
@@ -325,18 +325,18 @@
...
 				}
 
 				&:hover {
 					+.overlay {
-						left: 0;
 						opacity: 1;
-						backdrop-filter: blur(6px);
-						transition: backdrop-filter ease-in-out 1s 0.1s;
+                        left: 0%;
+                        backdrop-filter: brightness(30%);
```
flynnduism, line 338: Changing the `blur` to `brightness` is still problematic for browsers that struggle with the CSS `filter` effects - mostly Safari on iOS which will often crash and fail to load the page. 


**Accessibility / perf feedback:**
The `filter` and `backdrop-filter` classes should be avoided - you could do a less demanding `background-color: rgba(0,0,0,0.3);` (dark) or `background-color: rgba(255,255,255,0.3);` (light) approach and improve browser handling.

**Design feedback**
This :hover overlay effect is a nice-to-have flourish, and IMO it feels a bit heavy handed. It's a bit subjective, but a full page overlay is more typical with something like a modal than a dropdown menu. If it's taking over the page then ideally it's brief and subtle (not high contrast). 

The preview of this PR looks fine on dark mode, but is too high-contrast on light mode IMO.

itowlson:
Incidentally, if addressing this is going to turn into a marathon, then I'm happy to leave it as is.  It's a cosmetic niggle and I only brought it up again because I was going through old issues seeing if any could be closed.

karthik2804:
@itowlson  - It is a small change and I think it would be good to tackle in terms of browser compatibility and also website performance. 

Here is an updated preview link
https://fermyon-developer-twia6xje.fermyon.app/

cc: @flynnduism

itowlson:
Thanks!  The dark mode now feels gentle and subtly muted.  However, the light mode seems to _brighten_ the page, rather than dimming it.  Is that intentional?  For me left it looking the page looking a bit washed out - subjective I know though!

karthik2804:
@itowlson that was intentional as if not, we run again into the same contrast problem but I can take another stab at it.

flynnduism (approved):
I like how this edits looks and feels - subtle, doesn't distract or takeover the page as much.

There’s a slight a difference in the implementation between the light modes, but IMO it doesn’t matter, individually they both feel nice.

PR #1123, opened 2024-01-03
increase font weight on links

### LINKED ISSUE #1122: Are links too muted?
I added a link into the manifest reference, but when I went to check it, I found it very hard to see:

image

Could we make links a bit more distinct please?

karthik2804:
fix #1122 

This PR increases the font weight of the links in the content to make them a bit more bold and distinguishable.

image
image


itowlson (approved):
This LGTM - definitely makes it easier to notice "oh that bit of text is different, perhaps it is of interest", without feeling cluttery when links are mixed in with normal text.

PR #1070, opened 2023-12-04
[WIP] Add Kubernetes Scenario docs

voigt:
Content must go through a pre-merge checklist.

## Pre-Merge Content Checklist

This documentation has been checked to ensure that:

- [X] The `title, `template`, and `date` are all set
- [X] Does this PR have a new menu item (anywhere in `templates/*.hbs` files) that points to a document `.md` that is set to publish in the future? If so please only publish the `.md` and `.hbs` changes in real-time (otherwise there will be a menu item pointing to a `.md` file that does not exist)
- [X] File does not use CRLF, but uses plain LF (hint: use `cat -ve <filename> | grep '^M' | wc -l` and expect 0 as a result) 
- [X] Has passed [`bart check`](https://developer.fermyon.com/bartholomew/quickstart)
- [X] Has been manually tested by running in Spin/Bartholomew (hint: use `PREVIEW_MODE=1` and run `npm run styles` to update styling)
- [X] Headings are using Title Case
- [X] Code blocks have the programming language set to properly highlight syntax and the proper copy directive
- [X] Have tested with `npm run test` and resolved all errors
- [X] Relates to an existing (potentially outdated) blog article? If so please add URL in blog to point to this content.

PR #1134, opened 2024-01-11
update template to fix active page

karthik2804:
Switching from highlighting active elements based on a partial match to needing a full match for `/kubernetes` as that is a substring in other URLs. 

If we prefer to update everything to need a full match to keep up with consistency, I can do that as well. But I think I will make a follow-up PR cleaning up templates a little bit.

### DIFF
--- a/templates/spin_sidebar_v2.hbs
+++ b/templates/spin_sidebar_v2.hbs
@@ -251,17 +251,21 @@
                         Kubernetes
                 </label>
                 <ul class="menu-list accordion-menu-item-content">
-                        <li><a {{#if (active_project request.spin-full-url "/spin/v2/kubernetes" )}} class="active"
-                                        {{/if}} href="{{site.info.base_url}}/spin/v2/kubernetes">Spin on Kubernetes</a>
+                        <li><a {{#if (active_content request.spin-path-info "/kubernetes" )}} class="active" {{/if}}
+                                        href="{{site.info.base_url}}/spin/v2/kubernetes">Spin on Kubernetes</a>
                         </li>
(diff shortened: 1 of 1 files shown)


karthik2804:
Will update the PR to have wider clean ups and make this more consistent across all templates.

templates/spin_sidebar_v2.hbs
```diff
@@ -251,17 +251,21 @@
                         Kubernetes
                 </label>
                 <ul class="menu-list accordion-menu-item-content">
-                        <li><a {{#if (active_project request.spin-full-url "/spin/v2/kubernetes" )}} class="active"
-                                        {{/if}} href="{{site.info.base_url}}/spin/v2/kubernetes">Spin on Kubernetes</a>
+                        <li><a {{#if (active_content request.spin-path-info "/kubernetes" )}} class="active" {{/if}}
```
tpmccallum, line 254: Hi @karthik2804 
My apologies; I don't understand the changes here `active_content` vs `active_project` and `path-info` vs `full-url`.
When I preview the documentation side by side with a checkout of this PR they seem to have the same behaviour. I feel like I am totally missing something here.

karthik2804: HI @tpmccallum This address the issue pointed out in this [PR comment](https://github.com/fermyon/developer/pull/1070#issuecomment-1840843426)

tpmccallum (approved):

tpmccallum:
Will be sure to check this is fully resolved as part of the v3_ToC work.
Looks like there are other PR associated with this; but will test to be sure.

karthik2804:
It appears that this PR is no longer required.

PR #1137, opened 2024-01-12
add reo.dev script

karthik2804:


flynnduism (changes requested):
Looks good overall - for the Klaro cookie blocking-until-consent to work properly a change is needed to cintwnt_top.hbs

Example from .[com here](https://github.com/fermyon/fermyon.com/blob/main/templates%2Fcontent_bottom.hbs#L117)

1. the <script> tag should have an ID set that matches the name defined in the Klaro config JS in the content_bottom.hbs

2. The src property is usually changed to data-src so nothing is fetched upon pageload, and Klaro initiates a swap later when the consent is given. In this configuration, it looks like it's lazy loading itself - not sure exactly how, but the script probably needs a tweak to ensure the third external source is not fetched before Klaro can determine consent

karthik2804:
@flynnduism looks like it needs to be just `data-name`; for this kind of script, just setting the `data-type` to `text/plain` should be enough to stop it from parsing and executing and klaro should switch it out once the cookie consent is over. I could test it locally and see that it only loaded if the cookie consent was established

ref: https://klaro.org/docs/getting-started

flynnduism:
Tested locally, thanks @karthik2804 this works great.

flynnduism (approved):

PR #1151, opened 2024-01-19
Update registry-tutorial.md

sohanmaheshwar:
Adding:

1. Section for Deploy to Cloud
2. Callout saying that the package should be public (as its private by default) The error message we throw doesn't indicate this.

### DIFF
--- a/content/spin/v2/registry-tutorial.md
+++ b/content/spin/v2/registry-tutorial.md
@@ -45,7 +45,7 @@ To use a GHCR instance, you need to set up authentication. Follow [these steps](
 <!-- @nocpy -->
 
 ```bash
-> Login Succeed
+> Login Succeeded
 ```
 
 ## Push a Spin App to GHCR
@@ -100,6 +100,16 @@ Lastly, let's run this Spin application:
  ```bash
 $ spin up -f ghcr.io/USERNAME/spin-react-fullstack:v1
 ```
+
+## Deploy a Spin App from GHCR
+
+You can deploy a Spin app from GHCR by passing the GHCR URL as a query parameter. To do so, just use the URL in this format: 
+
+```
+https://cloud.fermyon.com/deploy?artifact=<GHCR-URL-HERE>&utm_source=hub&utm_medium=btn&utm_campaign=deploy-btn
+```
+
+> **Note:** You need to sign up for a Fermyon cloud account for this. Also, ensure that your package visilibty is set to 'Public' as the default package visibility on Github is private
 
 ## Conclusion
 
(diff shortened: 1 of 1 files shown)


content/spin/v2/registry-tutorial.md
```diff
@@ -101,6 +101,16 @@ Lastly, let's run this Spin application:
 $ spin up -f ghcr.io/USERNAME/spin-react-fullstack:v1
 ```
 
+## Deploy a Spin App from GHCR
+
+You can deploy a Spin app from GHCR by passing the GHCR URL as a query parameter. To do so, just use the URL in this format:
```
vdice, line 106: Note that users can do the same directly from the cli, for any remote app reference, including from GHCR, docker.io, etc,  eg `spin deploy -f <remote-reference-here>`.

This brings the same disclaimer that you've added below re: needing a cloud account -- though in this case, the reference actually _can_ be private, as long as the user is authenticated locally to the registry in question, since the cli will pull it down prior to publishing it to Cloud's internal registry.

I wonder if supplying the CLI command here might be a more natural fit?  Or maybe in addition?

tpmccallum (commented):
Hi @sohanmaheshwar 
This looks great, if Vaughn's suggestion of including `spin deploy -f` is added in this tutorial then LGTM.

content/spin/v2/registry-tutorial.md
itowlson: Yeah that is the internal mechanism behind the Deploy to Cloud button and I don't think we want users using it.  If we want to offer direct deploy then we should add a UX for it (or modify `spin deploy` to use the `/deploy` endpoint instead of pushing to the Cloud registry).  In the meantime definitely change to `spin deploy -f <reference>`.

sohanmaheshwar:
@vdice and @itowlson - made the changes based on your suggestions re: using only `spin deploy -f`
Have not mentioned the Deploy to Cloud button for now.

itowlson (approved):

vdice (approved):
Minor suggestions, otherwise LGTM!

content/spin/v2/registry-tutorial.md
```diff
@@ -101,6 +102,13 @@ Lastly, let's run this Spin application:
 $ spin up -f ghcr.io/USERNAME/spin-react-fullstack:v1
 ```
 
+## Deploy a Spin App from GHCR (or any registry)
+
+You can deploy a Spin app from any registry such as GHCR or Docker using the CLI command `spin deploy -f <remote-reference-here>`
```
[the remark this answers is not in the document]
sohanmaheshwar: Thanks @vdice !

tpmccallum (approved):

PR #1157, opened 2024-01-23
Fix broken link (vault variable test moved in Spin repo)

tpmccallum:

### DIFF
--- a/content/spin/v2/dynamic-configuration.md
+++ b/content/spin/v2/dynamic-configuration.md
@@ -94,7 +94,7 @@ $ vault kv put secret/password value="test_password"
 $ vault kv get secret/password
 ```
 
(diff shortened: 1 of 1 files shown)


itowlson (approved):

rylev (commented):
@tpmccallum unfortunately the vault test is no longer able to be built independently outside of our testing infrastructure. If we want users to be able to run this independently we should probably make an example in the examples directory instead of relying on tests.

tpmccallum:
I see that the Spin version in that test is `v1` also.
https://github.com/fermyon/spin/blob/main/tests/testcases/vault-variables-test/spin.toml#L1
Is that ok e.g. is that backward compatible or could/should we create a new `v2` version?
And also create a vault variable test application in the examples folder using Spin 2 template?

tpmccallum:
This has been updated at 
https://github.com/fermyon/spin/pull/2241/
and
https://github.com/fermyon/developer/pull/1162

Issue #1155, opened 2024-01-23
Add the ability to link to individual fields within the spin manifest documentation


endocrimes:
When trying to share spin documentation with people it would often be useful to link to a specific field in the manifest (e.g when mentioning `allowed_outbound_hosts`) - unfortunately you can only currently link to the overall _table_ for a section, a la https://developer.fermyon.com/spin/v2/manifest-reference#the-component-table.

itowlson:
@karthik2804 Is there a way to automate this in Bartholomew (like it does for headings) or is it just down to sticking `<a>` elements in by hand?

karthik2804:
Of the top of my head I can think of a few ways to achieve this. 

- like you pointed out, we can manually add link tags
- we can add a shortcode that automatically grabs the next element and create an anchor 
- if it is limited to just this page or just a couple other we can write some JavaScript to deal with it automatically on the frontend.

VamshiReddy02:
Hey @karthik2804 , can I work on this issue?

karthik2804:
@VamshiReddy02 sure but I am not sure we have a preferred approach here yet.

itowlson:
I have no opinion on the approach I just want @endocrimes to be happy

@karthik2804 decision or leave it up to @VamshiReddy02?

karthik2804:
I would be in favor of doing something with an annotating comment before the table and just writing some JS to generate the anchor links on the front end. 

@VamshiReddy02 let me know if you have other ideas or if you need help with the above approach.

PR #1165, opened 2024-01-25
Adding anchor-link to individual fields within the spin documentation

VamshiReddy02:
Fixes: #1155 

### DIFF
--- a/static/js/main.js
+++ b/static/js/main.js
@@ -1107,26 +1107,40 @@ const addCopyButtons = (clipboard)=>{
         pre.appendChild(button);
     });
 };
-const addAnchorLinks = ()=>{
-    document.querySelectorAll(".content h1, .content h2, .content h3, .content h4").forEach((heading)=>{
-        let id = heading.innerText.toLowerCase().replace(/[`~!@#$%^&*()_|+=?;:'",.<>\{\}\[\]\\\/]/gi, "").replace(/ +/g, "-");
-        heading.setAttribute("id", id);
-        heading.classList.add("heading-anchor");
-        let anchor = document.createElement("a");
-        anchor.className = "anchor-link";
-        anchor.href = "#" + id;
+const addAnchorLinks = () => {
+    const elementsToProcess = document.querySelectorAll(".content h1, .content h2, .content h3, .content h4, .content tr");
+    const idMap = new Map(); // Map to store the counts of each ID
+    elementsToProcess.forEach(element => {
+        let id;
+        if (element.tagName.toLowerCase() === 'tr') {
+            const rowContent = element.cells[0].textContent.trim();
+            id = `${rowContent.replace(/[`~!@#$%^&*()_|+=?;:'",.<>\{\}\[\]\\\/]/gi, '').replace(/ +/g, '-')}`;
+        } else {
+            id = element.innerText.toLowerCase().replace(/[`~!@#$%^&*()_|+=?;:'",.<>\{\}\[\]\\\/]/gi, '').replace(/ +/g, '-');
+        }
+        let count = idMap.get(id) || 0;
+        count++;
+        idMap.set(id, count);
+        if (count > 1) {
+            id += `${count}`;
+        }
+        element.setAttribute("id", id);
+        element.classList.add("heading-anchor");
+        let anchor = document.createElement('a');
+        anchor.className = 'anchor-link';
+        anchor.href = '#' + id;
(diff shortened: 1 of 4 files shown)


static/js/main.js
```diff
@@ -1107,26 +1107,40 @@ const addCopyButtons = (clipboard)=>{
...
-        anchor.className = "anchor-link";
-        anchor.href = "#" + id;
+const addAnchorLinks = () => {
+    const elementsToProcess = document.querySelectorAll(".content h1, .content h2, .content h3, .content h4, .content tr");
+    const idMap = new Map(); // Map to store the counts of each ID
+    elementsToProcess.forEach(element => {
+        let id;
+        if (element.tagName.toLowerCase() === 'tr') {
+            const rowContent = element.cells[0].textContent.trim();
+            id = `${rowContent.replace(/[`~!@#$%^&*()_|+=?;:'",.<>\{\}\[\]\\\/]/gi, '').replace(/ +/g, '-')}`;
```
karthik2804, line 1117: if two cells have the same content, this will lead to an ID collision (which can happen if we have 2 tables). We should devise a way here to generate a consistent unique ID for the table.

static/sass/styles.scss
```diff
@@ -420,6 +427,10 @@ Importing color and theme updates for the brand 1.5 refresh
 		}
 	}
 
+	thead {
+		background-color: #0a1931 !important;
+	}
```
karthik2804, line 432:
```suggestion
	tr {
		overflow-x: hidden;

		a.anchor-link {
			position: absolute;
		}
	}
```
we do not want to have an extra column in the table because it would look odd, so we can use `position: absolute` to get around it. The background-color becomes unnecessary once the anchor link is absolutely positioned

karthik2804 (changes requested):
This seems to be a good approach overall and gets really close to what we would like to have. I have a couple of concerns/suggestions regarding the styling of the anchor tags and ID generation for the anchor links.

Thanks for working on this!

static/js/main.js
VamshiReddy02: I think the code below ensures that the generated IDs for anchor links are unique, especially when there are cells in a table with the same content, right?
` let count = idMap.get(id) || 0;
        count++;
        idMap.set(id, count);
        if (count > 1) {
            id += ${count};
        }
`

static/sass/styles.scss
VamshiReddy02: Got it!

static/js/main.js
karthik2804: Yep I think that should be fine. My apologies I missed that.

itowlson: Are the IDs generated by a runtime counter going to be stable over time?

VamshiReddy02: I don't think so because If the content of the page changes, new elements are added, or existing elements are removed or modified, the counter may result in different IDs being generated.

itowlson: In that case I'm afraid this might need a rethink.  The purpose of this is to enable other content to link to the right parts of the table, and it's not generally convenient or even possible to update that other content if an anchor changes.  Sorry.

tpmccallum:
Hi @VamshiReddy02 
Is it possible for you to write some Javscript that will find the table row's closest parent heading with the class attribute of `heading-anchor`.
I think that combining the previous heading text and the table row text (to lower kebab case, say) will provide a unique enough value to be stable over time (as apposed to counting/incrementing number values).
What do you think?
Let me know if this makes sense and let me know if you would like a hand.
Thanks so much for your contribution 👍

VamshiReddy02:
Thank you @tpmccallum  for the suggestions, I'll try to implement for those suggestions.

tpmccallum (approved):
Tested this locally, works a treat, nicely done @VamshiReddy02

itowlson:
@tpmccallum @VamshiReddy02 I noticed in a recent article that uses one of the new links that we include the description as part of the link.  That means these anchors aren't going to be stable over time - as soon as we change a description, existing references will stop working, and users will be dropped in at the top of the page.  We need a solution that works not just at a point in time but across time.  Sorry for not noticing this when folks were working on it!

tpmccallum:
Thanks @itowlson - really great point. We may be able to update these row links to be more permanent.

@VamshiReddy02 I wonder if we can trim the URL back to just the row's `name`?

For example,  in the case of the application manifest reference table (say where we want to link to `allowed_http_hosts`) we could finish generating the hyperlink just after the `allowed_http_hosts` part. his would produce:

```
https://developer.fermyon.com/spin/v2/manifest-reference#spin-application-manifest-reference-allowed_http_hosts
```

Instead of:

```
https://developer.fermyon.com/spin/v2/manifest-reference#spin-application-manifest-reference-allowed_http_hosts-optional-array-of-strings-the-host-names-or-addresses-to-which-the-wasm-component-is-allowed-to-send-http-requests-this-is-retained-to-simplify-transition-from-the-version-1-manifest-new-applications-should-use-allow_outbound_hosts-instead-examplecom-localhost8081
```

itowlson:
My worry is that the `name` alone is ambiguous.  E.g. `variables` appears in more than one section.  @karthik2804 suggested nearest parent heading plus `name` (e.g. `manifest-reference#the-component-table-allowed_http_hosts`) would be unambiguous and likely to remain so.

@tpmccallum Would you prefer if I opened a new issue for this?

VamshiReddy02:
Hey @tpmccallum, Yes, I can trim the URL as you suggested. Additionally, the point raised by @itowlson is also valid. I can give a try to what @karthik2804 suggested.

tpmccallum:
Thanks @itowlson,
Yes, I agree with @karthik2804's idea. This was the suggestion proposed above i.e. "write some Javscript that will find the table row's closest parent heading"

Screenshot 2024-04-03 at 21 42 11

@VamshiReddy02, if you could trim the tail of the URL to be just the nearest parent heading plus that row's `name`, that would be greatly appreciated. You may run into a spot of bother. I tried for a few minutes and found that it was tricky to successfully parse the page's DOM and extract the elements and attributes that I wanted (namely being able to land on the closest parent heading). I hope you have great success. I have opened a new issue here with a little more detail https://github.com/fermyon/developer/issues/1235.

Of course please feel free to open a new PR and we will take a look as soon as you are ready.

karthik2804:
Something like this might just work (Have not tested)
```
function findNearestHeader(element) {
    var currentElement = element.parentNode;

    // Traverse up the DOM tree until a header element is found or no parent exists
    while (currentElement) {
      if (currentElement.nodeName.startsWith('h')) {
        // If the current element is a header element, return it
        return currentElement;
      }
      currentElement = currentElement.parentNode;
    }

    // If no header element is found in the ancestor elements, return null
    return null;
  }
```

tpmccallum:
Thanks @karthik2804 I will take this code suggestion over to the new [issue](https://github.com/fermyon/developer/issues/1235) for preservation.

PR #1202, opened 2024-02-29
update python router sample to componentize-py

karthik2804:
I spoke to @technosophos about switching the source of the example to the fermyon org and we are in agreement.

### DIFF
--- a/content/api/hub/sample_router.md
+++ b/content/api/hub/sample_router.md
@@ -5,7 +5,7 @@ content-type = "text/html"
 tags = ["python", "http", "routing"]
 
 [extra]
-author = "technosophos"
+author = "fermyon"
 type = "hub_document"
 category = "Sample"
 language = "Python"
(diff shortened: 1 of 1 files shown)


content/api/hub/sample_router.md
```diff
@@ -5,24 +5,24 @@ content-type = "text/html"
...
+author = "fermyon"
 type = "hub_document"
 category = "Sample"
 language = "Python"
 created_at = "2023-12-08T00:00:00Z"
 last_updated = "2023-12-08T00:00:00Z"
 spin_version = ">v2.0"
 summary =  "A sample that shows how to do internal HTTP routing in a Spin Python application"
-url = "https://github.com/technosophos/example-router-py"
+url = "https://github.com/fermyon/spin-python-sdk/tree/main/examples/external-lib-example"
```
itowlson, line 16: Is this showing a router or showing how to use an external library?

content/api/hub/sample_router.md
```diff
@@ -5,24 +5,24 @@ content-type = "text/html"
...
 keywords = "router, python"
 
 ---
 
 This project showcases how to do internal HTTP routing in a Spin Python application. It is similar to [the Spin JS router](https://developer.fermyon.com/spin/v2/javascript-components#routing-in-a-component) as well as the Django router.
 
-If you are interested in extrnal routing (where the routing table is declared in the spin.toml), the [Application Structure guide](https://developer.fermyon.com/spin/v2/spin-application-structure) covers this, as does the [documentation for Spin configuration](https://github.com/technosophos/example-router-py#:~:text=the%20documentation%20for%20Spin%20configuration).
+If you are interested in extrnal routing (where the routing table is declared in the spin.toml), the [Application Structure guide](https://developer.fermyon.com/spin/v2/spin-application-structure) covers this.
```
itowlson, line 23:
```suggestion
If you are interested in external routing (where the routing table is declared in the spin.toml), the [Application Structure guide](https://developer.fermyon.com/spin/v2/spin-application-structure) covers this.
```

content/api/hub/sample_router.md
karthik2804: It does both. It uses an external library and how it can be used to route.

content/api/hub/sample_router.md
itowlson: Also a bit disconcerting that I'm looking at an application called `external-something` and it's being _contrasted_ with `external-something-else`

content/api/hub/sample_router.md
itowlson: Hmm, okay.  I am concerned that if someone comes to the repo looking for "that router example they showed" then it's going to be highly non-obvious where to find it, but maybe that's managed through a read-me or something.  And I guess is out of scope for the Hub card.

karthik2804: Not sure I follow, It redirects directly to the example.

itowlson: No, sorry, this is out of scope for the Hub card but is musing on someone going to the repo and looking for things (which, realistically, is what many people do).  They are thinking about "router" and they see this:

image

But I agree this is a discussion for the SDK repo not for the Hub card.

tpmccallum (approved):
Thanks - this works for me.

```bash
spin build --up
Building component test with `componentize-py -w spin-http componentize app -o app.wasm`
Component built successfully
Finished building all Spin components
Preparing Wasm modules is taking a few seconds...

Serving http://127.0.0.1:3000
Available Routes:
  test: http://127.0.0.1:3000 (wildcard)
```

```bash
curl -i 127.0.0.1:3000
HTTP/1.1 200 OK
content-type: text/plain
transfer-encoding: chunked
date: Mon, 04 Mar 2024 03:31:08 GMT

Hello World
```

PR #1204, opened 2024-03-04
Update Spin documentation to include version 2.3

tpmccallum:

✅ content/spin/v2/cli-reference.md
✅ content/spin/v2/triggers.md


content/spin/v2/triggers.md
```diff
@@ -13,9 +13,70 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/triggers.md"
 
 A Spin _trigger_ maps an event, such as an HTTP request or a Redis pub-sub message, to a component that handles that event.
 
-An application can contain multiple triggers, but they must all be the same type. For example, an application can contain triggers for multiple HTTP routes, or for multiple Redis pub-sub channels, but can't contain both an HTTP _and_ a Redis trigger.
+An application can contain multiple triggers. As of Spin 2.3 onward, a Spin application can use any combination of trigger types within a single application. For example, a single application can contain triggers for both HTTP routes and Redis pub-sub channels. 
+
+> Currently, only the HTTP template supports the `spin add` command. If you wish to use `spin add` to build an application with multiple triggers, **including a Redis trigger**, you should begin with the Redis trigger application. Then, `spin add` your HTTP trigger component(s) as demonstrated below. Please note that additional Redis trigger components cannot be added via `spin add` at this time.
```
itowlson, line 18: I explained this badly when we discussed it - there isn't a single "HTTP template," and the various HTTP templates don't _necessarily_ support being added - it's only I think Rust and Go and the "off the shelf" ones like the fileserver which do.  Here's a crack at a re-word.

```suggestion
> Not all templates support the `spin add` command. In particular, at the time of writing, none of the default Redis templates support being added to existing applications. Therefore, if you want to use `spin add` to build an application with both Redis and HTTP triggers, you should first create a Redis application, then use `spin add` to add HTTP triggers, as shown below. (You won't be able to add additiona Redis triggers this way; if you need those, you'll need to set them up manually for now.)
```

content/spin/v2/triggers.md
```diff
@@ -13,9 +13,70 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/triggers.md"
...
+
+Here is a quick example of creating an application with both HTTP and Redis triggers:
```
itowlson, line 20:
```suggestion
Here is an example of creating an application with both HTTP and Redis triggers:
```

content/spin/v2/triggers.md
```diff
@@ -13,9 +13,70 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/triggers.md"
...
+Redis channel: one
+# Change into to the application directory
+$ cd trigger-example 
+# Create HTTP trigger application
+$ spin add -t http-rust http-trigger-example  
+Description: A HTTP trigger example
+HTTP path: /...
+```
+
+The above `spin new` and `spin add` commands will automatically scaffold the following Spin manifest (`spin.toml` file):
```
itowlson, line 38: This is getting very long and intricate for the opening paragraphs of the page.  Consider moving this to a "Setting up Multiple Trigger Types" section further down, after the orientation material.

I'd also strongly suggest pruning the TOML down to what you want people to look at.  The current TOML block is about 30 lines, most of which are boilerplate like "authors" and "component.*.build"  or empty `allowed_outbound_hosts` lines.  These make it hard to the eye to home in on what's important or distinctive about this particular TOML file.

content/spin/v2/triggers.md
```diff
@@ -13,9 +13,70 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/triggers.md"
...
+[component.http-trigger-example]
+source = "http-trigger-example/target/wasm32-wasi/release/http_trigger_example.wasm"
+allowed_outbound_hosts = []
+[component.http-trigger-example.build]
+command = "cargo build --target wasm32-wasi --release"
+workdir = "http-trigger-example"
+watch = ["src/**/*.rs", "Cargo.toml"]
+```
+
+> If you're familiar with [Spin 1.x](/spin/manifest-reference-v1#the-trigger-table), note that [Spin 2 uses the term "trigger"](/spin/manifest-reference#the-trigger-table) to refer to each individual route or channel, rather than the trigger type. It's closer to the `[component.trigger]` usage than to the application trigger.
```
itowlson, line 77: This shouldn't be in the middle of a discussion of multi-triggering - it's general orientation material for people migrating from 1.x.

content/spin/v2/triggers.md
```diff
@@ -13,9 +13,70 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/triggers.md"
...
 
-> If you're familiar with Spin 1.x, note that Spin 2 uses the term "trigger" to refer to each individual route or channel, rather than the trigger type. It's closer to the `[component.trigger]` usage than to the application trigger.
+Except for the application's business logic that an application developer will implement, the above multi-trigger application example is operational and can be built and run using `spin build --up` command.
```
itowlson, line 79: This paragraph doesn't really seem to be saying anything (other than maybe "we promise you this really works") - I'd suggest removing it, or rephrasing to clarify its intent.

content/spin/v2/triggers.md
```diff
@@ -13,9 +13,70 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/triggers.md"
 
 A Spin _trigger_ maps an event, such as an HTTP request or a Redis pub-sub message, to a component that handles that event.
 
-An application can contain multiple triggers, but they must all be the same type. For example, an application can contain triggers for multiple HTTP routes, or for multiple Redis pub-sub channels, but can't contain both an HTTP _and_ a Redis trigger.
+An application can contain multiple triggers. As of Spin 2.3 onward, a Spin application can use any combination of trigger types within a single application. For example, a single application can contain triggers for both HTTP routes and Redis pub-sub channels.
```
itowlson, line 16: Incidentally, even in 2.3, multi-trigger-type is an experimental feature with some rough edges...

content/spin/v2/triggers.md
```diff
@@ -69,11 +130,11 @@ These ways of writing components achieve the same thing, so which should you cho
 Named components have the following advantages:
 
 * Reuse. If you want two triggers to behave in the same way, they can refer to the same named component. Remember this means they are not just handled by the same Wasm file, but with the same settings.
-* Named. If an error occurs, Spin can tell your the name of the component where the error happened. With inline components, Spin has to synthesize a name. This isn't a big deal in single-component apps, but make diagnostics harder in larger apps.
+* Named. If an error occurs, Spin can tell you the name of the component where the error happened. With inline components, Spin has to synthesize a name. This isn't a big deal in single-component apps but makes diagnostics harder in larger apps.
```
itowlson, line 133:
```suggestion
* Named. If an error occurs, Spin can tell you the name of the component where the error happened. With inline components, Spin has to synthesize a name. This isn't a big deal in single-component apps, but makes diagnostics harder in larger apps.
```

content/spin/v2/triggers.md
```diff
@@ -69,11 +130,11 @@ These ways of writing components achieve the same thing, so which should you cho
...
 
 Inline components have the following advantages:
 
 * Compact, especially when using inline table syntax.
 * One place to look. Both the trigger event and the handling details are always in the same piece of TOML.
 
-If you are not sure, or are not experienced, we recommend using named components at first, and adopting inline components as and when you find cases where you prefer them.
+If you are not sure or are not experienced, we recommend using named components at first and adopting inline components as and when you find cases where you prefer them.
```
itowlson, line 140: I found this easier to parse with the commas.

content/spin/v2/cli-reference.md
```diff
@@ -1850,4 +2583,22 @@ CLI commands have four phases that indicate levels of stability:
...
+| <code>spin build</code>                                                               | Stable       |
+| <code>spin new</code>                                                                 | Stable       |
+| <code>spin plugins</code>                                                             | Stable       |
+| <code>spin templates</code>                                                           | Stable       |
+| <code>spin up</code>                                                                  | Stable       |
+| <code>spin cloud</code>                                                               | Stabilizing  |
+| <code>spin registry</code>                                                            | Stabilizing  |
+| <code>spin doctor</code>                                                              | Experimental |
+| <code>spin watch</code>                                                               | Experimental |
+| <code>spin bindle</code>                                                              | Deprecated   |
```
itowlson, line 2600: Hmm.  I'm pretty sure `spin bindle` hasn't existed since 2.0.

itowlson (changes requested):
My main concern is to retain the structure of the triggers orientation material.  Putting all the multi-trigger caveats and TOML up front drowns the "this is what triggers are" orientation material in a large and very detailed exposition of a specific authoring scenario.  I'd like to see the details of multi-trigger app creation put under its own section, on the same level as "Triggers and Components" and below it.  Actually, thinking about it, there's a case for the "spin new / spin add" stuff being on the "Writing Applications" and/or "Structuring Applications" page, which are where we get into more detail about using templates and looking at manifests... very much open to thoughts here...?

content/spin/v2/triggers.md
```diff
@@ -13,9 +13,74 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/triggers.md"
 
 A Spin _trigger_ maps an event, such as an HTTP request or a Redis pub-sub message, to a component that handles that event.
 
-An application can contain multiple triggers, but they must all be the same type. For example, an application can contain triggers for multiple HTTP routes, or for multiple Redis pub-sub channels, but can't contain both an HTTP _and_ a Redis trigger.
+An application can contain multiple triggers.
 
-> If you're familiar with Spin 1.x, note that Spin 2 uses the term "trigger" to refer to each individual route or channel, rather than the trigger type. It's closer to the `[component.trigger]` usage than to the application trigger.
+In Spin 2.2 and earlier, all triggers must be of the same type.  For example, an application can contain triggers for multiple HTTP routes, or for multiple Redis pub-sub channels, but not both.
+
+in Spin 2.3 and later, an application can contain triggers of different types.  For example, a single application can serve HTTP on one or more routes, and at the same time subscribe to one or more Redis pub-sub channels.
```
tpmccallum, line 20:
```suggestion
In Spin 2.3 and later, an application can contain triggers of different types.  For example, a single application can serve HTTP on one or more routes, and at the same time subscribe to one or more Redis pub-sub channels. 
```

content/spin/v2/triggers.md
tpmccallum: Updates in next push.

content/spin/v2/triggers.md
tpmccallum: Removed this paragraph. ✅

content/spin/v2/triggers.md
tpmccallum: Updated in new push. Thanks @itowlson!

content/spin/v2/cli-reference.md
tpmccallum: Thanks, removed the Bindle reference.

content/spin/v2/triggers.md
tpmccallum: Move back to the original order in latest push.

content/spin/v2/triggers.md
```diff
@@ -69,11 +74,57 @@ These ways of writing components achieve the same thing, so which should you cho
...
+Redis channel: one
+# Change into to the application directory
+$ cd trigger-example 
+# Create HTTP trigger application
+$ spin add -t http-rust http-trigger-example  
+Description: A HTTP trigger example
+HTTP path: /...
+```
+
+The above `spin new` and `spin add` commands will automatically scaffold a Spin manifest (`spin.toml` file) similar to the following:
```
itowlson, line 110: Should we indicate more explicitly that this is only the "interesting" bit of the manifest, e.g.
 
```suggestion
The above `spin new` and `spin add` commands will scaffold a Spin manifest (`spin.toml` file) with the following triggers:
```

itowlson (approved):
Apart from a small, non-blocking suggestion, LGTM.  Nice one turning this around so quickly!

content/spin/v2/triggers.md
tpmccallum: Oooh, yes, I like that!

PR #1229, opened 2024-03-26
Service chaining

itowlson:
**This is part of Spin 2.4.**  It should not be merged until Spin 2.4 is released.

### DIFF
--- a/content/spin/v2/http-outbound.md
+++ b/content/spin/v2/http-outbound.md
@@ -8,7 +8,9 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/http-outbo
 ---
 - [Using HTTP From Applications](#using-http-from-applications)
 - [Granting HTTP Permissions to Components](#granting-http-permissions-to-components)
-  - [Making HTTP Requests Within an Application](#making-http-requests-within-an-application)
+- [Making HTTP Requests Within an Application](#making-http-requests-within-an-application)
+  - [Local Service Chaining](#local-service-chaining)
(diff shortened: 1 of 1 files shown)


content/spin/v2/http-outbound.md
```diff
@@ -170,13 +172,40 @@ The Wasm module can make HTTP requests _only_ to the specified hosts. If a port
...
+allowed_outbound_hosts = ["*://*.spin.internal"]
+```
+
+> Local service chaining is not currently supported on Fermyon Cloud.
+
+### Intra-Application HTTP Requests by Route
+
+To make an HTTP request to another route with your application, you can pass just the route as the URL. For example, if you make an outbound HTTP request to `/api/customers/`, Spin prepends the route with whatever host the application is running on. It also replaces the URL scheme (`http` or `https`) with the scheme of the current HTTP request. For example, if the application is running in the cloud, Spin changes `/api` to `https://.../api`.
+
+In this way of doing self-requests, the request undergoes normal HTTP processing once Spin has prepends the host. For example, in a cloud deployment, the request passes through the network, and potentially back in through a load balancer or other gateway. The benefit of this is that it allows load to be distributed across the environment, but it may count against your use of bandwidth.
```
vdice, line 208:
```suggestion
In this way of doing self-requests, the request undergoes normal HTTP processing once Spin has prepended the host. For example, in a cloud deployment, the request passes through the network, and potentially back in through a load balancer or other gateway. The benefit of this is that it allows load to be distributed across the environment, but it may count against your use of bandwidth.
```

vdice (approved):

PR #1244, opened 2024-04-05
fix(install): update install script to support more OS types

kate-goldenring:
Updates the install script to:
1. detect if the OS is using musl and should use the bin with statically linked deps
2. the $OSTYPE is just "linux" to support OpenSUSE OS (issue described in discord [here](https://discord.com/channels/926888690310053918/1220198778288603247/1220206645544615976)
3. Removes unnecessary wrapping of spin directory in array which makes the syntax unsupported by `sh`

### DIFF
--- a/downloads/install.sh
+++ b/downloads/install.sh
@@ -89,12 +89,28 @@ case $UNAME_ARC in
     ;;
 esac
 
+# If OS uses musl, set distinct OS type to ensure the binary
+# with statically linked libraries and dependencies is used
+if command -v ldd >/dev/null 2>&1 && [[ "$(ldd /bin/ls | grep -m1 'musl')" ]]; then
+    OSTYPE="linux-musl"
+fi
+
 case $OSTYPE in
 "linux-gnu"*)
     OS="linux"
+    STATIC="false"
     ;;
 "darwin"*)
     OS="macos"
+    STATIC="false"
+    ;;
+"linux-musl"*)
+    OS="linux"
+    STATIC="true"
+    ;;
+"linux"*)
+    OS="linux"
+    STATIC="false"
     ;;
 *)
     fancy_print 1 "The OSTYPE: ${OSTYPE} is not supported by this script."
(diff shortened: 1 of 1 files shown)


karthik2804:
We may also want to optionally handle the installation of plugins. The plugins installed on `linux-musl` will probably be broken as we do not distinguish between libc.

downloads/install.sh
```diff
@@ -89,12 +89,28 @@ case $UNAME_ARC in
...
+# If OS uses musl, set distinct OS type to ensure the binary
+# with statically linked libraries and dependencies is used
+if command -v ldd >/dev/null 2>&1 && [[ "$(ldd /bin/ls | grep -m1 'musl')" ]]; then
+    OSTYPE="linux-musl"
+fi
+
 case $OSTYPE in
 "linux-gnu"*)
     OS="linux"
+    STATIC="false"
```
lann, line 101: I think you can drop this `linux-gnu` block entirely now right? The `linux*` below should catch it.

kate-goldenring:
> We may also want to optionally handle the installation of plugins. The plugins installed on `linux-musl` will probably be broken as we do not distinguish between libc.

@karthik2804 to clarify, that means folks using a distro that uses musl cannot write JS spin apps because that requires a plugin right? Should we track this as an issue somewhere -- building plugins with statically linked deps?

downloads/install.sh
kate-goldenring: Good point. Removed it

karthik2804:
@kate-goldenring, Yeah - any of the plugins will not work. The JS stuff will be replaced soon once we move over but things like `cloud`, `kube` and the trigger-plugins will not. I am not sure building all plugins as statically linked binaries is the best approach here. We may want spin to gain knowledge and fetch appropriate plugins?

flynnduism:
@kate-goldenring @karthik2804 

As we prepare to migrate spin docs to a CNCF org, can we review and resolve this PR?

kate-goldenring:
@karthik2804 how about i update this to not install plugins for a musl os?

karthik2804:
@kate-goldenring That sounds reasonable as users will have to build their own static release.

kate-goldenring:
@karthik2804 realizing i already made that change. Just rebased -- this is ready for review

karthik2804 (approved):

Issue #1253, opened 2024-04-15
Document using async functions with the Rust http handler


mikkelhegn:
Based on this question in Discord: https://discord.com/channels/926888690310053918/950022897160839248/1229360799429165066 I was not able to find information about using async functions in the Rust http handler in the docs.

tpmccallum:
Thanks @mikkelhegn 
New PR here https://github.com/fermyon/developer/pull/1255

PR #1255, opened 2024-04-16
Add improved search for `async` in Rust documentation.

tpmccallum:
Fixes https://github.com/fermyon/developer/issues/1253

This has already been documented, but it is not easily found. Therefore, this PR has been created to hoist that page to the surface. And still leave the other search results intact.

Before:

Screenshot 2024-04-16 at 11 19 59

After:

Screenshot 2024-04-16 at 11 19 42

This takes the user to the example shown below:

Screenshot 2024-04-16 at 11 20 48

### DIFF
--- a/content/spin/v2/rust-components.md
+++ b/content/spin/v2/rust-components.md
@@ -330,7 +330,7 @@ mod api {
 }
 ```
 
-> For further reference, see the [Spin SDK HTTP router](https://fermyon.github.io/rust-docs/spin/main/spin_sdk/http/struct.Router.html).
+> The Spin SDK HTTP router has additional async methods. For example, the `get_async` and `post_async` methods register an async handler at the path for the HTTP GET and HTTP POST methods. For further reference, see the [Spin SDK HTTP router](https://fermyon.github.io/rust-docs/spin/main/spin_sdk/http/struct.Router.html).
 
 ## Storing Data in Redis From Rust Components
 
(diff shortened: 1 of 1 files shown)


content/spin/v2/rust-components.md
```diff
@@ -328,7 +330,7 @@ mod api {
 }
 ```
 
-> For further reference, see the [Spin SDK HTTP router](https://fermyon.github.io/rust-docs/spin/main/spin_sdk/http/struct.Router.html).
+> The Spin SDK HTTP router has additional async methods. For example, the `get_async` and `post_async` methods register an async handler at the path for the HTTP GET and HTTP POST methods. For further reference, see the [Spin SDK HTTP router](https://fermyon.github.io/rust-docs/spin/main/spin_sdk/http/struct.Router.html).
```
itowlson, line 333: I would prefer not to make the "for further reference" seem specific to the async case.  What if we broke this into two paragraphs, e.g.

```suggestion
Handlers within a `Router` can be sync or async. Use `Router`'s "plain" methods (e.g. `get`, `post`) to assign synchronous handlers, and its "async" methods (e.g. `get_async`, `post_async`) for asynchronous handlers.  You can mix sync and async handlers in the same `Router`, and can use `handle` or `handle_async` to invoke `Router` processing, regardless of whether invididual handlers are sync or async.

> For further reference, see the [Spin SDK HTTP router](https://fermyon.github.io/rust-docs/spin/main/spin_sdk/http/struct.Router.html).
```

I also added some detail about the relationship between sync and async stuff that the original questioner was seeking clarification about.

itowlson (approved):

PR #1263, opened 2024-04-26
Adding Spin 2.5 documentation

tpmccallum:

### DIFF
--- a/content/spin/v2/cli-reference.md
+++ b/content/spin/v2/cli-reference.md
@@ -40,6 +40,7 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/cli-refere
 - [spin templates list](#spin-templates-list)
 - [spin templates uninstall](#spin-templates-uninstall)
 - [spin templates upgrade](#spin-templates-upgrade)
+- [spin test](#spin-test)
 - [spin up](#spin-up)
   - [spin up (HTTP Trigger)](#spin-up-http-trigger)
 - [spin watch](#spin-watch)
@@ -3353,6 +3354,36 @@ OPTIONS:
 {{ blockEnd }}
 
 **Note:** For additional information, please see the [Managing Templates](/spin/managing-templates) and/or [Creating Templates](/spin/template-authoring) sections of the documentation.
+
+<!-- markdownlint-disable-next-line titlecase-rule -->
+## spin test
+
+{{ startTab "v2.5"}}
+
+<!-- @selectiveCpy -->
+
+```console
+$ spin test --help
+
+By default `spin-test` will invoke the `run` subcommand.
+
+Usage: test [COMMAND]
+
+Commands:
+  run   Run a test suite against a Spin application
+  help  Print this message or the help of the given subcommand(s)
+
+Options:
+  -h, --help
+          Print help (see a summary with '-h')
+
+  -V, --version
+          Print version
+```
+
+{{ blockEnd }}
+
+{{ blockEnd }}
 
 <!-- markdownlint-disable-next-line titlecase-rule -->
 ## spin up
--- a/content/spin/v2/cli-reference.md
+++ b/content/spin/v2/cli-reference.md
@@ -200,6 +200,44 @@ OPTIONS:
 ```
 
 {{ blockEnd }}
+
+{{ startTab "v2.5"}}
+
+<!-- @selectiveCpy -->
+
+```console
+$ spin add --help
+spin-add 
+Scaffold a new component into an existing application
+
+USAGE:
+    spin add [OPTIONS] [NAME]
+
+ARGS:
+    <NAME>    The name of the new application or component
+
+OPTIONS:
+    -a, --accept-defaults              An optional argument that allows to skip prompts for the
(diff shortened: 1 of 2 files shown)


content/spin/v2/triggers.md
```diff
@@ -87,24 +87,22 @@ If you are not sure, or are not experienced, we recommend using named components
...
-> Note: Not all templates support the `spin add` command. In particular, at the time of writing, none of the default Redis templates support being added to existing applications. Therefore, if you want to use `spin add` to build an application with both Redis and HTTP triggers, you should first create a Redis application, then use `spin add` to add HTTP triggers, as shown below. (You won't be able to add additiona Redis triggers this way; if you need those, you'll need to set them up manually for now.)
-
 Here is an example of creating an application with both HTTP and Redis triggers:
 
 <!-- @nocpy -->
 
 ```bash
-# Start with a Redis trigger application
-$ spin new -t redis-rust trigger-example
+# Start with HTTP trigger application
```
itowlson, line 95: Good catch to update this - thanks!  However, I'd suggest following https://www.fermyon.com/blog/spin-application-structure and doing:

```
$ spin new -t http-empty trigger-example
$ cd trigger-example
$ spin add -t http-rust http-trigger-example
$ spin add -t redis-rust redis-trigger-example
```

This will give an evenly laid out directory structure instead of the HTTP code being in a top-level directory and the Redis code in a child one.  (However, it will have a knock-on effect on some of your example manifests, as the HTTP component will now have extra directories.)

content/spin/v2/triggers.md
```diff
@@ -115,16 +113,34 @@ The above `spin new` and `spin add` commands will scaffold a Spin manifest (`spi
 spin_manifest_version = 2
 
 [application]
-name = "trigger-example"
+name = "http-trigger-example"
+version = "0.1.0"
+authors = ["Your Name <your-name@example.com>"]
+description = "A HTTP trigger example"
```
itowlson, line 119: Maybe 

```suggestion
description = "An application that handles both HTTP requests and Redis messages"
```

?

content/spin/v2/cli-reference.md
```diff
@@ -707,6 +860,26 @@ OPTIONS:
...
+{{ blockEnd }}
+
+<!-- markdownlint-disable-next-line titlecase-rule -->
+## spin kube
+
+{ tabs "spin-version" }}
+
+{{ startTab "v2.5"}}
+
+The `spin kube` command is implemented by the [spin-plugin-kube](https://www.spinkube.dev/docs/spin-plugin-kube/reference/).
```
itowlson, line 879:
```suggestion
The `spin kube` command is implemented by the [`spin-kube` plugin](https://www.spinkube.dev/docs/spin-plugin-kube/reference/).
```

content/spin/v2/cli-reference.md
```diff
@@ -707,6 +860,26 @@ OPTIONS:
...
+{{ blockEnd }}
+
+{{ blockEnd }}
+
+<!-- markdownlint-disable-next-line titlecase-rule -->
+## spin kube
+
+{ tabs "spin-version" }}
+
+{{ startTab "v2.5"}}
```
itowlson, line 877: I don't think `spin kube` is tied to 2.5?

content/spin/v2/cli-reference.md
```diff
@@ -19,6 +19,7 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/cli-refere
 - [spin deploy](#spin-deploy)
 - [spin doctor](#spin-doctor)
 - [spin help](#spin-help)
+- [spin kube](#spin-kube)
```
itowlson, line 22: We don't usually list plugins in the command reference, except for `cloud` which is because it's installed by default.  If we are intending that `spin kube` is "default enough" to list here, should we do the same as for `spin cloud` and also list subcommands?

(For what it's worth, I would be entirely happy to drop the `cloud` subcommands here, and point people to the cloud plugin docs.   But whatever we decide our philosophy to be, it would be good to be consistent with it.)

itowlson (approved):
You will be astonished to learn that I have some suggestions and nitpicks.  But nothing blocking - thanks for working through all this.

I'm not sure if this is intended to cover all 2.5 changes?  It looks like granular route matching has landed in time so we'll need to write that up as well, but it can be a separate PR if that works better for you.  There is also private endpoints which I think is new, and I'm not sure if the OpenTelemetry support is new or was in 2.4?

content/spin/v2/triggers.md
tpmccallum: Great idea. I will update now.

content/spin/v2/cli-reference.md
tpmccallum: Listing the high-level commands and linking off to the official documentation, like we do with Cloud.

content/spin/v2/cli-reference.md
tpmccallum: I could back-date this for 2.4 also, but it is probably not relevant anymore since we are past that point.

itowlson:
`spin test` is a plugn isn't it?

PR #1265, opened 2024-04-30
Granular routing in Spin 2.5

tpmccallum:


tpmccallum:
Couple of questions for this one:

- is this the correct place for these docs? (Is this Rust specific?)
- what additional information is needed to satisfy the "granular" part of this functionality (given that the `:` variable segment has somewhat been covered off already in the preceding paragraph and code example)

content/spin/v2/rust-components.md
```diff
@@ -12,6 +12,7 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/rust-compo
 - [Redis Components](#redis-components)
 - [Sending Outbound HTTP Requests](#sending-outbound-http-requests)
 - [Routing in a Component](#routing-in-a-component)
+- [Granular Routing](#granular-routing)
```
itowlson, line 15: Oh sorry this isn't a Rust thing it's at the Spin manifest level.  It should be an extra paragraph in https://developer.fermyon.com/spin/v2/http-trigger#http-trigger-routes (with a mention in https://developer.fermyon.com/spin/v2/http-trigger#additional-request-information for how to access the match info).  I can send a PR for it if you like?

content/spin/v2/rust-components.md
```diff
@@ -334,6 +335,10 @@ Handlers within a `Router` can be sync or async. Use `Router`'s "plain" methods
 
 > For further reference, see the [Spin SDK HTTP router](https://fermyon.github.io/rust-docs/spin/main/spin_sdk/http/struct.Router.html).
 
+## Granular Routing
+
+Granular routing allows for more precise control over the request-handling process. Based on the [routefinder](https://crates.io/crates/routefinder) crate, granular routing in Spin allows users to define dynamic routes where certain parts of the URL can change based on the input provided. A colon (`:`) precedes a variable segment in a URL. This segment captures part of the URL as a variable, which can be used in your route handler. For example, `/goodbye/:planet`. In this route, `:planet` is a variable segment. When a user accesses a URL like `/goodbye/mars`, the router matches this to the `/goodbye/:planet` pattern, and `mars` is captured as the value of `planet`. The `RouteMatch` for this request would include the handler associated with the pattern and a parameter map containing `planet: mars`; this parameter map is what your application uses to handle and subsequently process the request.
```
itowlson, line 340: RouteMatch is an implementation detail of the trigger: it's not visible to guests.  Guests still get the standard HTTP API, with matches delivered by headers in a similar way to current `spin-path-info`.

itowlson (changes requested):
My apologies, I should have taken some time with you to explain what the new feature was - it's actually a HTTP trigger feature not a Rust SDK feature.  Sorry for not giving you better info (and I realise the PR doesn't say much about the developer experience either...).

content/spin/v2/http-trigger.md
tpmccallum, line 1: Just wondering what to mention inside the "Additional Request Information" section next.

content/spin/v2/http-trigger.md
```diff
@@ -78,6 +78,8 @@ route = "/users/..."
 component = "user-manager"
 ```
 
+Granular routing in Spin allows users to define dynamic routes where certain parts of the URL can change based on the input provided. A colon (`:`) precedes a variable segment in a URL. This variable segment captures part of the URL as a variable, which can be used in your route handler. For example, `/goodbye/:planet`. In this route, `:planet` is a variable segment. When a user accesses a URL like `/goodbye/mars`, the router matches this to the `/goodbye/:planet` pattern, and `mars` is captured as the value of `planet`.
```
itowlson, line 81: I wouldn't use the term "granular routing" - from a user POV this isn't a distinct 'granular routing' feature, it's just another kind of wildcard.  How about we change line 67 to something like:

-----

You can use wildcards to match 'patterns' of routes.  Spin supports two kinds of wildcard: _single-segment_ wildcards and _trailing_ wildcards.

A single-segment wildcard uses the syntax `:name`, where `name` is a name that identifies the wildcard.  Such a wildcard will match only a single segment of a path, and allows further matching on segments beyond it.  For example, `/users/:userid/edit` matches `/users/1/edit` and `/users/alice/edit`, but does not match `/users`, `/users/1`, or `/users/1/edit/cart`.

A _trailing wildcard_ uses the syntax `/...` and matches the given route and any route under it. (then continues as current para and example)

You can combine single-segment and trailing wildcards in the same route, e.g. `/users/:userid/cart/...`.

See Additional Request Information _(link)_ for how to access wildcard values from your component.

-----

(I've suggested changing your planets example to something that meshes a bit more with the running example of users and shops, and perhaps motivates the concept a bit more.  If you prefer to keep the planets then no worries.)

content/spin/v2/http-trigger.md
```diff
@@ -344,16 +346,16 @@ As well as any headers passed by the client, Spin sets several headers on the re
 > In the following table, the examples suppose that:
 > * Spin is listening on `example.com:8080`
 > * The application `base` is `/shop`
-> * The trigger `route` is `/users/...`
-> * The request is to `https://example.com:8080/shop/users/1/edit?theme=pink`
+> * The trigger `route` is `/users/:userid/cart/...`
+> * The request is to `https://example.com:8080/shop/users/1/cart/items/3`
```
itowlson, line 350: Worth keeping the query string to show how it's affected?

content/spin/v2/http-trigger.md
```diff
@@ -344,16 +346,16 @@ As well as any headers passed by the client, Spin sets several headers on the re
...
-| `spin-path-info`             | The request path relative to the component route (including any base) | `/1/edit` |
-| `spin-matched-route`         | The part of the request path that was matched by the route (including the base and wildcard indicator if present) | `/shop/users/...` |
-| `spin-raw-component-route`   | The component route pattern matched, as written in the component manifest (that is, _excluding_ the base, but including the wildcard indicator if present) | `/users/...` |
-| `spin-component-route`       | The component route pattern matched, _excluding_ any wildcard indicator | `/users` |
+| `spin-full-url`              | The full URL of the request. This includes full host and scheme information. | `https://example.com:8080/shop/users/1/cart/items/3` |
+| `spin-path-info`             | The request path relative to the component route (including any base) | `/items/3` |
+| `spin-matched-route`         | The part of the request path that was matched by the route (including the base and wildcard indicator if present) | `/shop/users/:userid/cart/...` |
```
itowlson, line 356: If that's the value then we might need to reword the description - this is no longer part of the request path, but rather the _trigger route_.

itowlson (changes requested):
I'd suggest integrating this more directly into the discussion of routes and wildcards instead of appending it as a sort of distinct feature.  I've left a rather lengthy suggestion I'm afraid - hope it's useful.

We should also mention this in the manifest reference I think (https://developer.fermyon.com/spin/v2/manifest-reference#additional-fields-for-triggerhttp-tables-route), as part of the route description - you may already be doing this in a different PR as it will otherwise conflict with the private endpoints change, not sure!

content/spin/v2/http-trigger.md
tpmccallum: Updating in the next push

content/spin/v2/http-trigger.md
tpmccallum: Sure thing, adding in now

content/spin/v2/http-trigger.md
```diff
@@ -78,6 +79,12 @@ route = "/users/..."
 component = "user-manager"
 ```
 
+You can use wildcards to match 'patterns' of routes. Spin supports two kinds of wildcard: single-segment wildcards and trailing wildcards.
```
itowlson, line 82: These paragraphs (current lines 82-86) should replace the first two sentences of line 68, rather than appearing after the existing content.  That was why I didn't put the text as a GH suggestion - we have an existing discussion of wildcard routes and the new content should be part of that.

content/spin/v2/http-trigger.md
```diff
@@ -122,6 +129,20 @@ route = "/..."
...
+
+<!-- @nocpy -->
+
+```toml
+[[trigger.http]]
+route = { private = true }
+component = "internal"
+```
+
+Private endpoints still require the `allowed_outbound_hosts` to be set. See [local service chaining documentation](./http-outbound#local-service-chaining) for more information about configuring self-requests (where the request is passed in memory without ever leaving the Spin host process).
```
itowlson, line 144:
```suggestion
To access a private endpoint, use [local service chaining](./http-outbound#local-service-chaining) (where the request is passed in memory without ever leaving the Spin host process). Such calls still require the internal endpoint to be included in `allowed_outbound_hosts`.
```

content/spin/v2/manifest-reference.md
```diff
@@ -152,7 +152,9 @@ Each array entry contains a mix of common fields and trigger-specific fields.
 
 | Name                    | Required?  | Type        | Value    | Example   |
 |-------------------------|------------|-------------|----------|-----------|
-| `route`                 | Required   | String      | The route which this component handles. Requests to the route will cause the component to execute. This may be an exact route (`/example`), which matches on the given path, or a wildcard route indicated by the suffix `/...` (`/example/...`), which matches any route under this prefix. If two routes overlap, requests are directed to the matching route with the longest prefix - see [the HTTP trigger documentation](http-trigger) for details and examples. | `"/api/cart/..."` |
+| `route`                 | Required   | String or Table     | The route which this component handles. Requests to the route will cause the component to execute. | |
```
itowlson, line 155:
```suggestion
| `route`                 | Required   | String or table     | The route which this component handles. Requests to the route will cause the component to execute. | |
```

(just for consistency with existing occurrence)

content/spin/v2/manifest-reference.md
```diff
@@ -152,7 +152,9 @@ Each array entry contains a mix of common fields and trigger-specific fields.
...
+|                         |            | String     | This may be an exact route (`/example`), which matches on the given path, or a wildcard route indicated by the suffix `/...` (`/example/...`), which matches any route under this prefix. If two routes overlap, requests are directed to the matching route with the longest prefix - see [the HTTP trigger documentation](http-trigger) for details and examples. | `"/api/cart/..."` |
```
itowlson, line 156: Do you want to mention single-segment wildcards here, or make the wildcard stuff more generic e.g.

```suggestion
|                         |            | String     | This may be an exact route (`/example`), which matches only the given path, or may include wildcards (`/example/:id` or `/example/...`). If two routes overlap, requests are directed to the matching route with the longest prefix. See [the HTTP trigger documentation](http-trigger) for details and examples. | `"/api/cart/..."` |
```

itowlson (commented):
This is coming together - thanks!  I had some minor suggestions on the reference but the main thing is I confused matters by writing proposed wildcard text in a comment on a new section when I felt (and still feel) it should become part of the existing flow.  Unfortunately GH didn't (and still doesn't) let me put the suggestion where I think it should go!  What I really feel we should do is update the existing wildcard content (line 68) rather than adding new content after the end of that section.  Does that make sense?  Sorry for the miscommunication!

content/spin/v2/http-trigger.md
```diff
@@ -64,7 +65,7 @@ route = "/cart/checkout"
 component = "checkout"
 ```
 
-A _wildcard_ route matches the given route and any route under it.  A route is a wildcard if it ends in `/...`.  For example, `/users/...` matches `/users`, `/users/1`, `/users/1/edit`, and so on.  Any of these routes will run the mapped component.
+You can use wildcards to match 'patterns' of routes. Spin supports two kinds of wildcards: single-segment wildcards and trailing wildcards. A _wildcard_ route matches the given route and any route under it.  A route is a wildcard if it ends in `/...`.  For example, `/users/...` matches `/users`, `/users/1`, `/users/1/edit`, and so on.  Any of these routes will run the mapped component.
```
itowlson, line 68:
```suggestion
You can use wildcards to match 'patterns' of routes. Spin supports two kinds of wildcards: single-segment wildcards and trailing wildcards.

A single-segment wildcard uses the syntax `:name`, where `name` is a name that identifies the wildcard. Such a wildcard will match only a single segment of a path, and allows further matching on segments beyond it. For example, `/users/:userid/edit` matches `/users/1/edit` and `/users/alice/edit`, but does not match `/users`, `/users/1`, or `/users/1/edit/cart`.

A trailing wildcard uses the syntax `/...` and matches the given route and any route under it.  For example, `/users/...` matches `/users`, `/users/1`, `/users/1/edit`, and so on. Any of these routes will run the mapped component.
```

content/spin/v2/http-trigger.md
```diff
@@ -78,6 +79,10 @@ route = "/users/..."
 component = "user-manager"
 ```
 
+A single-segment wildcard uses the syntax `:name`, where `name` is a name that identifies the wildcard. Such a wildcard will match only a single segment of a path, and allows further matching on segments beyond it. For example, `/users/:userid/edit` matches `/users/1/edit` and `/users/alice/edit`, but does not match `/users`, `/users/1`, or `/users/1/edit/cart`.
```
itowlson, line 82: remove if suggestion accepted (the suggestion moves this text to be part of the existing wildcard content)

content/spin/v2/http-trigger.md
```diff
@@ -78,6 +79,10 @@ route = "/users/..."
...
+
+A trailing wildcard uses the syntax `/...` and matches the given route and any route under it.
```
itowlson, line 84: remove (already covered by existing wildcard content, just slightly rephrased in the suggested revision to make a contrast withe single-segment wildcards)

content/spin/v2/http-trigger.md
```diff
@@ -344,16 +363,16 @@ As well as any headers passed by the client, Spin sets several headers on the re
...
 |------------------------------|----------------------|---------|
-| `spin-full-url`              | The full URL of the request. This includes full host and scheme information. | `https://example.com:8080/shop/users/1/edit?theme=pink` |
-| `spin-path-info`             | The request path relative to the component route (including any base) | `/1/edit` |
-| `spin-matched-route`         | The part of the request path that was matched by the route (including the base and wildcard indicator if present) | `/shop/users/...` |
-| `spin-raw-component-route`   | The component route pattern matched, as written in the component manifest (that is, _excluding_ the base, but including the wildcard indicator if present) | `/users/...` |
-| `spin-component-route`       | The component route pattern matched, _excluding_ any wildcard indicator | `/users` |
+| `spin-full-url`              | The full URL of the request. This includes full host and scheme information. | `https://example.com:8080/shop/users/1/cart/items/3/edit?theme=pink` |
+| `spin-path-info`             | The request path relative to the component route (including any base) | `/items/3` |
```
itowlson, line 372: I forgot that we ought to add a pattern for single-segment wildcard values (e.g. the `:userid` being `"1"`).  I'm not sure the best way to do it given that these will vary with the name of the wildcard.

itowlson (changes requested):
Sorry to keep at this - I'm clearly not expressing my suggestions well, and I apologise for that.  Fortunately at least with this one I can write what I'm suggesting in the right place so hopefully will make more sense now!

I realised there is also a new header - or a new 'pattern' of headers for single-segment wildcard values.  These are of the form `spin-path-match-<segment_name>` (e.g. `spin-path-match-userid`).  I'm not sure the best way to add this to the headers table because it's currently structured around exact strings rather than families/patterns - any ideas?  Thanks!

itowlson (approved):

PR #1266, opened 2024-04-30
Private endpoints in Spin 2.5

tpmccallum:

### DIFF
--- a/content/spin/v2/http-outbound.md
+++ b/content/spin/v2/http-outbound.md
@@ -12,7 +12,6 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/http-outbo
 - [Making HTTP Requests Within an Application](#making-http-requests-within-an-application)
   - [Local Service Chaining](#local-service-chaining)
   - [Intra-Application HTTP Requests by Route](#intra-application-http-requests-by-route)
-  - [Private Endpoints](#private-endpoints)
 
 Spin provides an interface for you to make outgoing HTTP requests.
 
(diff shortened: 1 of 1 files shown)


tpmccallum:
This would be better with a concrete example. Suggestions are welcome.

content/spin/v2/http-outbound.md
```diff
@@ -224,3 +225,7 @@ You must still grant permission by including `self` in `allowed_outbound_hosts`:
 ```toml
 allowed_outbound_hosts = ["http://self", "https://self"]
 ```
+
+### Private Endpoints
+
+Private endpoints represent internal microservices that are shielded from external network exposure. These endpoints do not have public HTTP routes and can only be accessed internally within the application's architecture.
```
mikkelhegn, line 231: It should be related to the service chaining feature. So an example could be that two components use this endpoint internally, through service chaining. Maybe it's not it's oiwn heading, but just a call-out in the service chaining section, that these components have endpoints, which are not exposed?

content/spin/v2/http-outbound.md
```diff
@@ -12,6 +12,7 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/http-outbo
 - [Making HTTP Requests Within an Application](#making-http-requests-within-an-application)
   - [Local Service Chaining](#local-service-chaining)
   - [Intra-Application HTTP Requests by Route](#intra-application-http-requests-by-route)
+  - [Private Endpoints](#private-endpoints)
```
itowlson, line 15: We should probably be mentioning these in the HTTP trigger page because that's where app developers will actually set up private endpoints.  Calling them is kind of already covered by service chaining, so we probably just need a mention in the service chaining docs about "this is the only way to call private endpoints" or something.

itowlson (changes requested):

content/spin/v2/http-outbound.md
tpmccallum: Thanks @mikkelhegn I have updated this page and this will be updated in the next push.

content/spin/v2/http-outbound.md
tpmccallum: Thanks @itowlson - updating this now.

itowlson (changes requested):
This really needs to be covered on the HTTP trigger page (and the manifest reference).  Service chaining users don't need to know anything about whether the endpoint is private or public, but we _do_ need to tell people how to set an endpoint to private (i.e. `route = { private = true }` in the manifest).  I can send a PR for this if it would help?

itowlson (approved):
Looks like we're covering the rest of this in #1265 so LGTM! :shipit:

PR #1270, opened 2024-05-03
Fix no copy code block and weird link

garikAsplund:
I think the example for `<!-- @nocpy -->` was off before--it had a copy block! Also there was a link that was kinda botched so fixed that.

When I run `cat -ve <filename> | grep '^M' | wc -l` I get 2 😞 


content/spin/v2/contributing-docs.md
```diff
@@ -175,9 +175,9 @@ hello
 
 The no copy annotation (`<!-- @nocpy -->`) precedes a code block where no copy and pasting of code is intended. If using the no copy attribute please still be sure to add the appropriate syntax highlighting to your code block (for display purposes). For example:
 
-`<!-- @nocpy -->`
+<!-- @nocpy -->
```
itowlson, line 178: I believe this line is intended to show the use of the `@nocpy` directive, so it's correct that it's backticked.

itowlson (changes requested):
Thanks for the catch on that link!  The `nocpy` change needs backing out as the backticks are intentional, otherwise should be good to go!

itowlson:
@garikAsplund Are the CR characters on lines you've changed?  Normally git takes care of this stuff - if you're on Windows and have autocrlf turned on then it shouldn't be an issue - and if you're on Linux or Mac then I'm not sure where they'd have come from!

content/spin/v2/contributing-docs.md
garikAsplund: But then you can copy from the example box, no? Seems strange to have it that way.

garikAsplund:
I'm on a Mac and was confused by it too! Hope it gets sorted out.

content/spin/v2/contributing-docs.md
itowlson: I guess we could do a backticked "nocpy" for display followed by an actual nocpy to suppress the copy button on the sample text?  I'm not too fussed about it but we do need to show what the contributor has to type as that is the whole point of the section _grin_

garikAsplund: Before: Screenshot 2024-05-02 at 10 11 35 PM
After: Screenshot 2024-05-02 at 10 11 07 PM

itowlson:
Huh.  I see _three_ matches for the grep and they are all "M at the beginning of a line".

@tpmccallum any idea what's going on with that check?

@karthik2804 does Bartholomew really still get mad about CRs? could we fix it at source and make this check go away?

content/spin/v2/contributing-docs.md
itowlson: @garikAsplund Yes. It is no longer showing the `<!-- nocpy -->` directive.  _The purpose of the example is the show the usage of the directive._  Please do not make it invisible.

garikAsplund: I think the directive is still there though--just in the paragraph--very much like in the previous selective copy example. The example _should be_ the code block. Note the icon difference between the two. It's strange to say "here's an example of a no copy block" and then be able to copy it.

itowlson: I understand that.  But the purpose of the example is to show _the source that the contributor would type, including the directive._

Perhaps an alternative way to approach this would be to code-quote the "Some generic code..." as well as the directive, so that it is all shown as "source code the contributor would type"?  Or to do that followed by the render with just the uncopyable code block a la:

> For example, if your Markdown file contains:
>
> _code block with directive and example_
>
> Then it will be rendered as:
>
> _the uncopyable block_

garikAsplund: I think that would be a little clearer than present and hopefully more consistent if the selective copy followed that format too.

karthik2804:
@itowlson I am not sure if Bartholomew gets mad about CRs but if that is the case, we can take a leaf out of this [PR](https://github.com/fermyon/bartholomew/pull/68) and just make the CRs vanish  (i.e) `doc.replace{"\r", ""}`

content/spin/v2/contributing-docs.md
garikAsplund: Don't mean to complicate or nitpick--I tried adding in the code block as an example and wasn't happy with it because rendering a to-be-copied code block _within_ a code block is insanely hard! `<code>` wasn't good, `` \` `` wasn't attractive, ZWJ has a good look but messes with user functionality in a hidden way, and on and on.

I still question whether there's a need for `<!-- @nocpy -->` as a standalone line especially if the example above doesn't have it.

Screenshot 2024-05-05 at 10 45 23 AM

tpmccallum:
Hi @garikAsplund 
Thanks for this contribution.
A few updates:
It looks like the `cat -ve <filename> | grep '^M' | wc -l` is actually matching on lines that start with upper-case `M`. This is not an issue. Apologies for the confusion. I will adjust the template so this is not a roadblock for others. 

I see you have tried really hard to incorporate the no copy in the example. This is a tough task. Perhaps an image will do the trick. Please see how the latest commit to this PR adds an image and resolves this meta issue.

I am happy with this PR now. Just touching in with you to see if you agree. Thanks again for catching this and creating the PR. Really appreciate your contribution.

tpmccallum:
Screenshot 2024-05-07 at 10 42 59

itowlson (approved):
Love this ingenious solution @tpmccallum - thanks for finding a way out of the impasse!

tpmccallum:
The ol' extra backtick trick
Screenshot 2024-05-07 at 11 06 57

Issue #1267, opened 2024-05-02
Mismatched Language: Python Code in TypeScript tab


VamshiReddy02:
While reviewing the [ Application variable](https://developer.fermyon.com/spin/v2/variables#using-variables-from-applications) documentation, I noticed that the codebase currently includes Python code within a TypeScript tab.

<img width="1440" alt="Screenshot 2024-05-02 at 11 49 26 AM" src="https://github.com/fermyon/developer/assets/99033623/3aab98ab-3d8d-45f2-84ed-c5413b279aa5">

PR #1271, opened 2024-05-03
Replacing python code with typescript in spin docs

VamshiReddy02:
fixes : #1267


tpmccallum (approved):
Thanks @VamshiReddy02

PR #1286, opened 2024-05-13
Adding cron documentation

tpmccallum:

- [x] File does not use CRLF, but uses plain LF (hint: use `cat -ve <filename> | grep $'\r' | wc -l` and expect 0 as a result)

### DIFF
--- a/content/spin/v2/cron-trigger.md
+++ b/content/spin/v2/cron-trigger.md
@@ -1,0 +1,90 @@
+title = "The Spin Cron Trigger"
+template = "spin_main"
+date = "2023-11-04T00:00:01Z"
+enable_shortcodes = true
+[extra]
+url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/cron-trigger.md"
+
+---
+
+Spin has built-in support for creating and running components driven by time. This page covers Spin options that are specific to the Spin Cron Trigger.
+
+The [experimental Cron trigger for Spin](https://github.com/fermyon/spin-trigger-cron) allows you to write and deploy workloads, in a similar fashion to the Cron command-line utility. Let's look at how to set this up and deploy a web application that is driven by time.
+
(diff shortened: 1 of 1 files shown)


content/spin/v2/triggers.md
```diff
@@ -147,4 +154,92 @@ allowed_outbound_hosts = []
...
+[1715640449] Hello from a cron component
+[1715640451] Hello from a cron component
+[1715640453] Hello from a cron component
+[1715640455] Hello from a cron component
+[1715640457] Hello from a cron component
+```
+
+As we can see from the above output, our application is now running and executing the function every two seconds without the need for any incoming requests or any intervention from users or other machines.
+
+If you would like to learn more about using the Spin Cron Trigger, please check out [the Spin Cron Trigger blog post](https://www.fermyon.com/blog/spin-cron-trigger) and the [Spin Cron Trigger GitHub Repository](https://github.com/fermyon/spin-trigger-cron).
```
itowlson, line 262:
```suggestion
If you would like to learn more about using the Spin Cron Trigger, please check out [the Spin Cron Trigger blog post](https://www.fermyon.com/blog/spin-cron-trigger) and the [Spin Cron Trigger GitHub repository](https://github.com/fermyon/spin-trigger-cron).
```

itowlson (approved):
There are some links from when this was a separate page that need to be fixed (but are routine to do).  I'd also strongly prefer to talk about "on a schedule" rather than "based on time" (in particular, "elapsed time" is not right), and have a few other suggestions, but am happy to defer all of this stuff to a tweak pass if you want to knock this out.

I'd also really like to see this on its own page, because to my mind is skews the triggers home page, which I see as being a landing page more than a reference for every trigger.  But again we can definitely leave that for a tweak pass.

Thanks as ever for working through all this and - also as ever - apologies for the graceless nitpicking!

Issue #1298, opened 2024-05-22
Update Developer Documentation to Include Python Version Requirement


kedaarsridhar20:
**Issue:**

The current [Spin Quickstart Guide](https://developer.fermyon.com/spin/v2/quickstart) does not mention that a specific version of Python is required for the setup and use of Spin when using Python. Specifically, the componentize-py package does not work unless Python version 3.10 or later is installed. This omission can lead to installation issues for developers following the guide who have older versions of Python.

**Proposed Change:**

Add a step in the "Install the Prerequisites" section of the Spin Quickstart Guide to instruct users to install Python 3.10 or later. The updated section should include the following information:

> Prerequisite: Install Python 3.10 or later
> 
> Ensure that you have Python 3.10 or later installed on your system. You can check your Python version by running:
> 
> ```
> python3 --version
> ```
> If you do not have Python 3.10 or later, you can install it by following the instructions [here](https://www.python.org/downloads/).

Rationale:

Including this prerequisite will help ensure that developers have the correct environment setup, preventing errors related to incompatible Python versions. This will improve the overall user experience and reduce setup-related issues.

Additional Information:

For more context on the requirement for Python 3.10 or later, please see the [componentize-py documentation](https://pypi.org/project/componentize-py/) which specifies the version compatibility.

PR #1299, opened 2024-05-23
Update documentation to clarify Python version required

tpmccallum:

Fixes https://github.com/fermyon/developer/issues/1298

- [x] File does not use CRLF, but uses plain LF (hint: use `cat -ve <filename> | grep $'\r' | wc -l` and expect 0 as a result)

### DIFF
--- a/content/spin/v2/ai-sentiment-analysis-api-tutorial.md
+++ b/content/spin/v2/ai-sentiment-analysis-api-tutorial.md
@@ -9,6 +9,7 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/ai-sentime
 - [Tutorial Prerequisites](#tutorial-prerequisites)
   - [Spin](#spin)
   - [Dependencies](#dependencies)
+- [Prerequisite](#prerequisite)
 - [Licenses](#licenses)
 - [Serverless AI Inferencing With Spin Applications](#serverless-ai-inferencing-with-spin-applications)
   - [Creating a New Spin Application](#creating-a-new-spin-application)
@@ -62,6 +63,16 @@ $ spin templates install --git https://github.com/fermyon/spin-js-sdk --upgrade
 ```
 
 **Python**
+
+## Prerequisite
+
+Ensure that you have Python 3.10 or later installed on your system. You can check your Python version by running:
+
+```bash
+python3 --version
+```
+
+If you do not have Python 3.10 or later, you can install it by following the instructions [here](https://www.python.org/downloads/).
 
 Some of the Serverless AI examples are written using Python. To enable Serverless AI functionality via Python, please ensure you have the latest Python template installed:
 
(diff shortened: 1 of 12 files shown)


content/spin/v2/ai-sentiment-analysis-api-tutorial.md
```diff
@@ -63,6 +64,16 @@ $ spin templates install --git https://github.com/fermyon/spin-js-sdk --upgrade
 
 **Python**
 
+## Prerequisite
```
itowlson, line 67: This is already in a "Tutorial Prerequisites" section so seems not necessary to repeat?  But if you do, don't headingify it because this implies that the "Tutorial Prequisites" section ends at `**Python**, and that the text after it, including the TinyGo section, is part of something new - basically don't use anything that adds a higher level of structure/navigation than the `**Python**` sub-sub-heading.

content/spin/v2/key-value-store-tutorial.md
```diff
@@ -72,6 +73,17 @@ $ spin new -t http-ts spin-key-value
 
 {{ startTab "Python" }}
 
+## Prerequisite
```
itowlson, line 76: This interacts uncomfortably with the "we'll start from a template using the following commands:" text that introduces the tab block.  The other blocks follow on from that, by starting with the template install command.

The best way to avoid this would be to move this paragraph to the existing "Tutorial Prerequisites" section.

If you strongly wish to have it here, you could put it _after_ the `spin new` command so that the "introductory para to tab content" flow is preserved.

content/spin/v2/key-value-store-tutorial.md
```diff
@@ -9,6 +9,7 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/key-value-
 - [Key Value Store With Spin Applications](#key-value-store-with-spin-applications)
 - [Tutorial Prerequisites](#tutorial-prerequisites)
 - [Creating a New Spin Application](#creating-a-new-spin-application)
+- [Prerequisite](#prerequisite)
```
itowlson, line 12: This is very contextless in the TOC.  Maybe it would be better not to use an actual heading for this.

content/spin/v2/quickstart.md
```diff
@@ -132,6 +132,14 @@ Installing template http-ts...
 
 {{ startTab "Python" }}
 
+Ensure that you have Python 3.10 or later installed on your system. You can check your Python version by running:
```
itowlson, line 135: This is in a section called "Install a Template".  It would fit more naturally in the "Install the Tools" section which is where we talk about things like having the right Rust bits or installing TinyGo.

content/spin/v2/writing-apps.md
```diff
@@ -211,6 +211,14 @@ HTTP path: /...
 
 {{ startTab "Python"}}
 
+Ensure that you have Python 3.10 or later installed on your system. You can check your Python version by running:
```
itowlson, line 214: This isn't particularly associated with the template - someone picking up an existing application to maintain might never use the template but would still need this info.  To me, it fits more naturally in the "Setting up for `spin build`" section on the build page, which is where we talk about required tools and tool configuration.

It might also be good to mention it in a build troubleshooting section.

itowlson (changes requested):
The content is fine but I'm a bit doubtful about some of the places it's been injected, especially when it's been given a H2 heading that causes it to enter the page TOC.  Could we revisit the positioning of the content with a slightly zoomed out view of the context please?  Sorry!

itowlson (approved):
LGTM - thanks!

PR #1305, opened 2024-06-03
Update sqlite-api-guide.md

tpmccallum:

### DIFF
--- a/content/spin/v2/sqlite-api-guide.md
+++ b/content/spin/v2/sqlite-api-guide.md
@@ -236,7 +236,7 @@ A convenience function `sqlite.Open()` is provided to create a database connecti
 
 {{ blockEnd }}
 
-> Using `COMMIT;` in SQLite is crucial as it finalizes a transaction, ensuring that all changes made during the transaction are saved permanently to the database. If you do not use `COMMIT,` you will get undesirable side effects in your database.
+> Using `COMMIT;` on `INSERT`, `UPDATE` and `DELETE` statements in SQLite is crucial as it finalizes a transaction, ensuring that all changes made during the transaction are saved permanently to the database. If you do not use `COMMIT;` you will get undesirable side effects in your database.
 
 ## Preparing an SQLite Database
 
--- a/content/spin/v2/sqlite-api-guide.md
+++ b/content/spin/v2/sqlite-api-guide.md
@@ -236,6 +236,8 @@ A convenience function `sqlite.Open()` is provided to create a database connecti
 
 {{ blockEnd }}
 
+> Using `COMMIT;` in SQLite is crucial as it finalizes a transaction, ensuring that all changes made during the transaction are saved permanently to the database. If you do not use `COMMIT,` you will get undesirable side effects in your database.
+
 ## Preparing an SQLite Database
 
 Although Spin provides SQLite as a built-in database, SQLite still needs you to create its tables.  In most cases, the most convenient way to do this is to use the `spin up --sqlite` option to run whatever SQL statements you need before your application starts.  This is typically used to create or alter tables, but can be used for whatever other maintenance or troubleshooting tasks you need.


content/spin/v2/sqlite-api-guide.md
```diff
@@ -236,6 +236,8 @@ A convenience function `sqlite.Open()` is provided to create a database connecti
 
 {{ blockEnd }}
 
+> Using `COMMIT;` on `INSERT`, `UPDATE` and `DELETE` statements in SQLite is crucial as it finalizes a transaction, ensuring that all changes made during the transaction are saved permanently to the database. If you do not use `COMMIT;` you will get undesirable side effects in your database.
```
itowlson, line 239: This is only for transactions, isn't it?  Or are you saying that Spin is doing something under the covers that creates a requirement for `COMMIT` even when there is no transaction?

tpmccallum: Yes, I found that Spin can re-run statements that modify data again during the next statement (if the original statement was not immediately followed by a `COMMIT;`).
I did not investigate further, just noticed Spin double-adding records and the `COMMIT;` fixed the issue at the time.

tpmccallum: I will try and reproduce today. BTW, I was using the Rust (so not sure if this behaviour is across all SQLite language implementations)

itowlson: If that's happening then it's a bug in Spin and we should fix at source.  Users should not, as far as I know, have to COMMIT unless a transaction is in play.  Happy to be corrected of course!

tpmccallum: Just discovered what was happening, with the duplicate content.

If there is a statement executing inside a HTTP component and the event is triggered via a browser (localhost:3000) as apposed to via CLI (curl localhost:3000) the duplicate happens because the function is executed twice:

```bash
tpmccallum@tpmccallum:~/Fermyon/sqlite-example$ spin build --up --sqlite @migration.sql
Building component sqlite-example with `cargo build --target wasm32-wasi --release`
   Compiling sqlite-example v0.1.0 (/home/tpmccallum/Fermyon/sqlite-example)
    Finished `release` profile [optimized] target(s) in 0.84s
Finished building all Spin components
Storing default SQLite data to ".spin/sqlite_db.db"

Serving http://127.0.0.1:3000
Available Routes:
  sqlite-example: http://127.0.0.1:3000 (wildcard)
```

When accessing via the browser the following happens:

```bash
Handling one request only to Some(HeaderValue { inner: String("http://127.0.0.1:3000/") })
Epoch time: 1719464512
Handling one request only to Some(HeaderValue { inner: String("http://127.0.0.1:3000/favicon.ico") })
Epoch time: 1719464512
```

tpmccallum: Note the second line with the `favicon.ico`.

tpmccallum: Ended up using some logic in the code to deal with the second (unwanted) request that the browser makes when the browser is searching for a `favicon`:

```rust
use spin_sdk::http::{IntoResponse, Request, Response};
use spin_sdk::http_component;
use spin_sdk::sqlite::{Connection, QueryResult, Row as OtherRow, Value};
use std::fs::read;
use std::time::{SystemTime, UNIX_EPOCH};

#[http_component]
fn handle_sqlite_example(req: Request) -> anyhow::Result<impl IntoResponse> {
    println!("Handling request to {:?}", req.header("spin-full-url"));
    if req
        .header("spin-path-info")
        .expect("No URL path info")
        .as_str()
        == Some("/favicon.ico")
    {
        return Ok(Response::builder()
            .status(404)
            .header("content-type", "text/plain")
            .body("Not found")
            .build());
    } else {
        let connection = Connection::open_default()?;
        let epoch_time = SystemTime::now().duration_since(UNIX_EPOCH)?.as_secs();
        println!("Epoch time: {}", epoch_time);
        let query = format!(
            "INSERT INTO TestTable (epoch_time) VALUES ({}); COMMIT;",
            epoch_time
        );
        connection.execute(&query, &[])?;
        Ok(Response::builder()
            .status(200)
            .header("content-type", "text/plain")
            .body("Performed database statement")
            .build())
    }
}

```

tpmccallum:
It appears that the `COMMIT` in SQLite is automatic. I will close this now that the double execution has been diagnosed and resolved.

PR #1323, opened 2024-06-20
Make linking consistent for upcoming migration

tpmccallum:
There is a mix of absolute and relative links to the developer.fermyon.com site.
This PR removes the absolute links and uses relative links (that take into account the original link before modifying, the redirects that are already in place i.e. `spin/` -> `spin/v2/` and the hierarchy of the `api/hub`, `spin/v1`, `spin/v2` and `cloud/changelog` directories which all require a `../../` if linking to a different root path.

These all resolve (just a quick mention that we are not able to link to the `api/hub` in a relative manner. That path does not resolve so we need to use the absolute URL in two or three cases).


content/cloud/pricing-and-billing.md
```diff
@@ -52,9 +52,9 @@ Fermyon charges the tax mandated by the local taxing authority. Your bill will i
 
 No, at this time Fermyon Cloud only accepts reoccurring payments. If at any point you decide you no longer want to continue with your paid plan, you are eligible to cancel your subscription. 
 
-**What happens if there's an issue with my payment method during a reoccurring charge?**
+**What happens if there's an issue with my payment method **during a **recurring** charge?****
```
itowlson, line 55: The nested bolds don't do anything.  Not sure what you're aiming for with the emphasis-upon-emphasis?  The original (with your wording tweak) seems fine to me

```suggestion
**What happens if there's an issue with my payment method during a recurring charge?**
```

content/cloud/pricing-and-billing.md
```diff
@@ -116,19 +116,19 @@ If these do not meet your needs, either email <a href="mailto:sales@fermyon.com"
...
+Fermyon will notify you via email when you reach 75%, 90%, and 100% usage for your limits. In the future we will be adding Fermyon Cloud UI notifications as well as the ability to configure alerts.
 
 **What happens when I reach my quota limits?**
 
-You'll receive an email from <a href="mailto:support@fermyon.com">support@fermyon.com</a> asking to establish a plan to pull your resources back into the appropriate limits for your existing plan, or alternatively upgrade your plan. In the future, resources will be thorttled at 100% utilization of quota limit for Starter Plan users, and Growth Users will be able to select between overrage charges or throtting at 100% utilization of quota limits. 
+You'll receive an email from <a href="mailto:support@fermyon.com">support@fermyon.com</a> asking to establish a plan to pull your resources back into the appropriate limits for your existing plan, or alternatively upgrade your plan. In the future, resources will be throttled at 100% utilization of the quota limit for Starter Plan users, and Growth Users will be able to select between coverage of charges or throttling at 100% utilization of quota limits.
```
itowlson, line 123: Pretty sure "overrage charges" is intended here, i.e. "charges for going over"

```suggestion
You'll receive an email from <a href="mailto:support@fermyon.com">support@fermyon.com</a> asking to establish a plan to pull your resources back into the appropriate limits for your existing plan, or alternatively upgrade your plan. In the future, resources will be throttled at 100% utilization of the quota limit for Starter Plan users, and Growth Users will be able to select between overrage charges or throttling at 100% utilization of quota limits. 
```

content/spin/v2/key-value-store-tutorial.md
```diff
@@ -522,6 +522,6 @@ Congratulations, you have a Spin application and associated Key Value store runn
 
 ## Next Steps
 
-* Explore the contents of your Key Value store with the [Key Value Store Explorer template](https://developer.fermyon.com/hub/preview/template_kv_explorer)
+* Explore the contents of your Key Value store with the [Key Value Store Explorer template](../../api/hub/preview/template_kv_explorer)
```
itowlson, line 525: This differs from the previous path by adding `api` on the front.  For me locally this causes it to 404.

itowlson (changes requested):
Couple of small things - one where wording has been changed in a way that changes the meaning (and it affects money so I'd be wary of doing so without socialising it), and one which introduces an accidental 404.  Otherwise LGTM and I'm stoked to see it!

content/spin/v2/key-value-store-tutorial.md
itowlson: I am curious why the link checker does not pick up on this one...

tpmccallum: Yeah, the `api/hub` baffle me.
I think we can't link to the home page of the hub relatively. It will be the one exception. 
I have made this one absolute to get around the 404 issue.
Perhaps @karthik2804 will know why we are not able to link to `api/hub` index page using relative path.

karthik2804: There are 2 sets of pages - one is `api/hub/...` which serves JSON data for the front-end (this should not be linked anywhere or site mapped). The other is what is served from the front-end which is `hub/...`. Therefore the URL here needs to be `/hub/preview/template_kv/explorer` or if you want relative URL, it is `../../hub/preview/template_kv_explorer`. 

The preview page comes from the `vue` app and is not part of the content files.

tpmccallum: Thanks, Karthik,
So if I understand it correctly there is no way to link to the https://developer.fermyon.com/hub (home page of the hub) relatively from within the GitHub markdown files. We will have to leave these `hub` related links as absolute links (I think there is only about 3 or 4 of them anyway).
This is ready for approval if you agree @itowlson - thanks so much!

itowlson: I am confused.  `../../hub/preview/template_kv_explorer` works for me (that is, it brings up the KV explorer Hub page), and seems to be what Karthik is suggesting.  Are you seeing something different @tpmccallum?

tpmccallum: Yeah, the issue is we can't link to the hub's landing page using relative links.
For example, there are instances in the docs where the hub's home page < https://github.com/fermyon/developer/pull/1323/files#diff-5c25aabf95d9789251ab01457e99ea55bf967e0d7e716f69208a64aa2817b61bL181 > is absolute and I wish we could link to the hub's home page using a relative link.

tpmccallum: There are only 3 links to the top-level landing page of the hub.
image
So don't want this to block the work. But it would be great to have that relative link ability, like we do with the other areas of the site.

tpmccallum: Oooh, might have found a loophole. 
If I use `/index` at the end of a relative link to the hub (i.e. `../hub/index`), it routes back to `/hub` (the user gets what they want), and the local broken link tests (npm run test) pass. I will try and push these changes and see if the build passes.

itowlson: > Linking to actual items is ok; the landing page is the issue.

I am still confused: don't we _want_ to link to an individual item here?  Sorry for being slow.

tpmccallum: All good Ivan, I will try and explain as best I can ...

Why link to the Spin Up Hub landing page?
The Python page, for example, is trying to link to the Spin Up Hub (in general) so users can dive in and use the filters to search for Python-related items. 

Why not just link to the Python section of the Spin Up Hub? 
When using the filters to browse different languages etc. the Spin Up Hub URL stays exactly the same. So we can never link to just Python-related things in the hub.

What types of links can we offer to users then?
There are only two types:
- A niche/specific link that only shows a single item i.e. [the Python HTTP Template](https://developer.fermyon.com/hub/preview/template_http_python) 
- [the landing page](https://developer.fermyon.com/hub)
There is no in between.

What was the original issue?
Sometimes (as is the case with this Python page) we just want to mention that a user can go to the Spin Up Hub at < https://developer.fermyon.com/hub > and use the filters to search for things. The Spin Up Hub works in such a way that using relative links in the documentation markdown files i.e. `../hub` did not pass the broken link checker in the build.

How has that been resolved?
It turns out that using `../hub/index` instead of just `../hub` manages to get past the broken link checker issue. We can now have 100% relative links in the documentation. 

Hope this makes sense.

itowlson (approved):
Thanks Tim!

tpmccallum:
Thanks @itowlson - sorry for the confusion.

PR #1324, opened 2024-06-21
Automate CLI documentation task

tpmccallum:
Adding toolkit to automate documentation tasks moving forward.


toolkit/README.md
vdice, line 1: Can we add or link to a section for python newbs (or less frequent users) about how to set things up to run the scripts?  (eg not sure if setting up a venv is necessary, pip install, etc)

Acting as the naive user here, this is what happens when I just jump in and start running the scripts:
```
$ python3 update_spin_cli_reference.py
Traceback (most recent call last):
  File "/Users/vdice/code/github.com/fermyon/developer/toolkit/update_spin_cli_reference.py", line 3, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
```

toolkit/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Usage
+
+Edit the version numbers in the file i.e. `2.4`, `2.5` etc. and run the script:
+
+```bash
+python3 update_spin_cli_reference.py
+```
+
+The following `check_cli_output.py` script will tell you what the changes are (between versions) these scripts are designed to quickly kick off the edit, review, approve, and merge process at the time of release.
```
vdice, line 18: This line seems a little out of place since it references the `check_cli_output.py` script discussed in the next section...

toolkit/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+We:
+- create a variable for each version
+- create two new directories
+- copy the script into each directory (this makes the script's output identical - aside from actual changes)
+- install the specific version of Spin in each directory
+- run the script using that specific version
+- diff the results
+
+```bash
```
vdice, line 34: Do we want to include this bash script (or some form of it) to make it even easier for folks to run it all?  eg user might then just need to run `check-cli-diff.sh v2.5.0 v2.6.0` (the script or user may need to do some python setup...)

toolkit/README.md
```diff
@@ -0,0 +1,67 @@
...
+# Write Codeblocks
+
+The `update_spin_cli_reference.py` script will fetch the current Spin CLI reference document from the web and write a fresh copy of the file (with the new version of code blocks added) to your local disk. This new markdown is the start of a PR that contains CLI commands for the new version of Spin.
+
+## Usage
+
+Edit the version numbers in the file i.e. `2.4`, `2.5` etc. and run the script:
+
+```bash
+python3 update_spin_cli_reference.py
```
vdice, line 15: Possible to update the script such that the versions can be passed in as env vars or args?  (To avoid needing to edit the hard-coded versions in the script itself).

vdice (commented):
Initial review; haven't yet tested the scripts themselves.  Definitely look like some handy tools! 🧰

toolkit/README.md
tpmccallum: Thanks @vdice - the instructions now set up a venv and show how to install the dependencies also.

toolkit/README.md
tpmccallum: Thanks, I have moved it under the proper heading for clarity.

toolkit/README.md
tpmccallum: Thanks @vdice - that's a great idea - we now have a bash script that takes two arguments, does the whole job and the clean up.

tpmccallum:
Thanks @vdice - I just did a pretty major update on all of these and tested it out.
It should be good to go.
Thanks for the review feedback and new ideas; greatly appreciated.

toolkit/script_to_check_cli_outputs.sh
```diff
@@ -0,0 +1,24 @@
...
+  echo "Usage: $0 v2.6.0 v2.7.0"
+  exit 1
+fi
+SPIN_CURRENT=$1
+SPIN_LATEST=$2
+mkdir ~/$SPIN_CURRENT
+mkdir ~/$SPIN_LATEST
+cp -rp check_cli_output.py ~/$SPIN_CURRENT
+cp -rp check_cli_output.py ~/$SPIN_LATEST
+cd ~
```
vdice, line 13: What do you think about using `/tmp`(or even better, `mktemp`) instead of `~`/`$HOME`?

toolkit/README.md
```diff
@@ -0,0 +1,81 @@
...
+Run the script using the `vX.X.X` format. Please note, the Spin versions you use must exist as a Spin release in the Spin GitHub repo.
+
+**Note**: Use three number characters `vX.X.X` i.e. `v2.6.0`:
+
+```bash
+# Assuming this repo is cloned to your home directory (as per the first example also)
+cd ~
+cd developer/toolkit
+# Make the script executable
+sudo chmod a+x script_to_check_cli_outputs.sh
```
vdice, line 70: We should be able to get rid of this step if you run `chmod +x script_to_check_cli_outputs.sh` and commit the result.

vdice (commented):
Thanks for the helpful updates @tpmccallum!  Just a few more suggestions and then we can start iterating on these tools.

toolkit/script_to_check_cli_outputs.sh
tpmccallum: Thanks @vdice - great idea. Using `mktemp` now.

tpmccallum:
Edited these files (squashed the commits) and tested it out locally again this afternoon (with the new updates).
Thanks for the suggestions and review @vdice - hopefully this is good to go now.

vdice (approved):
LGTM!

tpmccallum:
Thanks @vdice

PR #1355, opened 2024-07-30
update js content to v2 sdk

karthik2804:
This PR updates the documentation to be based on the upcoming release of the v2 of the JS SDK.

Todo:
- [ ] Check the links once the new docs are up and SDK is moved to `main`.
- [ ] update `serverless-ai-hello-world` once the example repo is updated.

### DIFF
--- a/content/spin/v2/build.md
+++ b/content/spin/v2/build.md
@@ -66,23 +66,17 @@ command = "cargo build --target wasm32-wasi --release"
 
 {{ startTab "TypeScript" }}
 
-For JavaScript and TypeScript applications, you must have the `js2wasm` Spin plugin installed:
+For JavaScript and TypeScript applications, you must have `node.js`.
 
-<!-- @selectiveCpy -->
 
-```bash
-$ spin plugins update
-$ spin plugins install js2wasm --yes
-```
-
-It's normally convenient to put the detailed build instructions in `package.json`:
+It's normally convenient to put the detailed build instructions in `package.json`, The build script calls out to [`webpack`](https://webpack.js.org/) and `j2w` which is a script provided by the `@fermyon/spin-sdk` package that utilizes [`ComponentizeJS`](https://github.com/bytecodealliance/ComponentizeJS). The build script looks like:
 
 <!-- @nocpy -->
 
 ```json
 {
   "scripts": {
-    "build": "npx webpack --mode=production && mkdir -p target && spin js2wasm -o target/spin-http-js.wasm dist/spin.js"
+   "build": "npx webpack --mode=production && npx mkdirp target && npx j2w -i dist.js -d combined-wit -n combined -o target/spin-http-js.wasm"
   }
 }
 ```
--- a/content/spin/v2/build.md
+++ b/content/spin/v2/build.md
@@ -69,17 +69,19 @@ command = "cargo build --target wasm32-wasi --release"
 For JavaScript and TypeScript applications, you must have [Node.js](https://nodejs.org).
 
 
-It's normally convenient to put the detailed build instructions in `package.json`. The build script calls out to [`webpack`](https://webpack.js.org/) and `j2w` which is a script provided by the `@fermyon/spin-sdk` package that utilizes [`ComponentizeJS`](https://github.com/bytecodealliance/ComponentizeJS). The build script looks like:
+It's normally convenient to put the detailed build instructions in `package.json`. The build script looks like:
 
 <!-- @nocpy -->
 
 ```json
 {
   "scripts": {
-   "build": "npx webpack --mode=production && npx mkdirp target && npx j2w -i dist.js -d combined-wit -n combined -o target/spin-http-js.wasm"
+    "build": "npx webpack --mode=production && npx mkdirp target && npx j2w -i dist.js -d combined-wit -n combined -o target/spin-http-js.wasm"
   }
 }
 ```
+
+{{ details "Parts of the build script" "The build script calls out to [`webpack`](https://webpack.js.org/) and `j2w` which is a script provided by the `@fermyon/spin-sdk` package that utilizes [`ComponentizeJS`](https://github.com/bytecodealliance/ComponentizeJS)."}}
 
 The build command can then call the NPM script:
 
--- a/content/spin/v2/build.md
+++ b/content/spin/v2/build.md
@@ -66,10 +66,10 @@ command = "cargo build --target wasm32-wasi --release"
 
 {{ startTab "TypeScript" }}
 
-For JavaScript and TypeScript applications, you must have `node.js`.
+For JavaScript and TypeScript applications, you must have [Node.js](https://nodejs.org).
 
 
-It's normally convenient to put the detailed build instructions in `package.json`, The build script calls out to [`webpack`](https://webpack.js.org/) and `j2w` which is a script provided by the `@fermyon/spin-sdk` package that utilizes [`ComponentizeJS`](https://github.com/bytecodealliance/ComponentizeJS). The build script looks like:
+It's normally convenient to put the detailed build instructions in `package.json`. The build script calls out to [`webpack`](https://webpack.js.org/) and `j2w` which is a script provided by the `@fermyon/spin-sdk` package that utilizes [`ComponentizeJS`](https://github.com/bytecodealliance/ComponentizeJS). The build script looks like:
 
 <!-- @nocpy -->
 
--- a/content/spin/v2/http-outbound.md
+++ b/content/spin/v2/http-outbound.md
@@ -102,8 +102,7 @@ const response = await fetch("https://example.com/users");
 
 **Notes**
 
-// TODO: Add example and update link
-You can find a complete example of using outbound HTTP in the JavaScript SDK repository on [GitHub](https://github.com/fermyon/spin-js-sdk/tree/main/examples/typescript/outbound_http)
+You can find a complete example of using outbound HTTP in the JavaScript SDK repository on [GitHub](https://github.com/fermyon/spin-js-sdk/tree/main/examples/common-patterns/outbound-http)
 
 **Note**: `fetch` currently only works when building for the HTTP trigger.  
 
--- a/content/spin/v2/http-outbound.md
+++ b/content/spin/v2/http-outbound.md
@@ -102,9 +102,10 @@ const response = await fetch("https://example.com/users");
 
 **Notes**
 
-* Although the underlying Spin interface is blocking, the `fetch` function is defined by JavaScript as async. You must await the response, but the request will always block, and the promise will resolve as soon as the request is returned.
+// TODO: Add example and update link
+You can find a complete example of using outbound HTTP in the JavaScript SDK repository on [GitHub](https://github.com/fermyon/spin-js-sdk/tree/main/examples/typescript/outbound_http)
 
-You can find a complete example of using outbound HTTP in the JavaScript SDK repository on GitHub ([TypeScript](https://github.com/fermyon/spin-js-sdk/tree/main/examples/typescript/outbound_http), [JavaScript](https://github.com/fermyon/spin-js-sdk/tree/main/examples/javascript/outbound-http)).
+**Note**: `fetch` currently only works when building for the HTTP trigger.  
 
 {{ blockEnd }}
 
--- a/content/spin/v2/http-trigger.md
+++ b/content/spin/v2/http-trigger.md
@@ -267,31 +267,14 @@ For a full Rust SDK reference, see the [Rust Spin SDK documentation](https://doc
 
 > [**Want to go straight to the reference documentation?**  Find it here.](https://fermyon.github.io/spin-js-sdk/)
 
-In JavaScript or TypeScript, the handler is identified by name.  It must be called `handleRequest`.  The way you declare it is slightly different between the two languages.
+In TypeScript, the user must a define a function named `handler` which the SDK attaches to the [`fetch` event listener](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent).
 
-In **JavaScript**, `handleRequest` is declared as `export async function`.  It takes a JavaScript [object representing the request](https://fermyon.github.io/spin-js-sdk/interfaces/HttpRequest.html), and returns a [response object](https://fermyon.github.io/spin-js-sdk/interfaces/HttpResponse.html).  The fields of these objects are exactly the same as in TypeScript:
+```ts
+import { ResponseBuilder } from "@fermyon/spin-sdk";
 
-```javascript
-export async function handleRequest(request) {
-    return {
-        status: 200,
-        headers: { "content-type": "text/plain" },
-        body: "Hello from JS-SDK"
-    }
-}
-```
-
-In **TypeScript**, `handleRequest` is declared as an `export const` of the [`HandleRequest` function type](https://fermyon.github.io/spin-js-sdk/types/HandleRequest.html) - that is, a function literal rather than a function declaration.  It takes a [`HttpRequest` object](https://fermyon.github.io/spin-js-sdk/interfaces/HttpRequest.html), and returns a [`HttpResponse` object](https://fermyon.github.io/spin-js-sdk/interfaces/HttpResponse.html), both defined in the `@fermyon/spin-sdk` package:
-
-```javascript
-import { HandleRequest, HttpRequest, HttpResponse} from "@fermyon/spin-sdk"
-
-export const handleRequest: HandleRequest = async function(request: HttpRequest): Promise<HttpResponse> {
-    return {
-      status: 200,
-        headers: { "content-type": "text/plain" },
-      body: "Hello from TS-SDK"
-    }
+export async function handler(req: Request, res: ResponseBuilder) {
+    console.log(req);
+    res.send("hello universe");
 }
 ```
 
--- a/content/spin/v2/http-trigger.md
+++ b/content/spin/v2/http-trigger.md
@@ -267,7 +267,9 @@ For a full Rust SDK reference, see the [Rust Spin SDK documentation](https://doc
 
 > [**Want to go straight to the reference documentation?**  Find it here.](https://fermyon.github.io/spin-js-sdk/)
 
-In TypeScript, the user must a define a function named `handler` which the SDK attaches to the [`fetch` event listener](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent).
+The user must a define a function named `handler` which the SDK attaches to the [`FetchEvent` listener](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent). Note that the incoming HTTP event is translated to a `FetchEvent`.
+
+The handler function takes in two arguments a [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) and a [ResponseBuilder](https://fermyon.github.io/spin-js-sdk/classes/ResponseBuilder.html)
 
 ```ts
 import { ResponseBuilder } from "@fermyon/spin-sdk";
--- a/content/spin/v2/javascript-components.md
+++ b/content/spin/v2/javascript-components.md
@@ -382,4 +382,4 @@ These are some of the suggested libraries that have been tested and confirmed to
 ## Caveats
 
 - All `spin-sdk` related functions and methods (like `Variables`, `Redis`, `Mysql`, `Pg`, `Kv` and `Sqlite`) can be called only inside the `handler` function. This includes `fetch`. Any attempts to use it outside the function will lead to an error. This is due to Wizer using only Wasmtime to execute the script at build time, which does not include any Spin SDK support.
-- No operation that involve handling private keys are suported. 
+- No crypto operation that involve handling private keys are supported. 
--- a/content/spin/v2/javascript-components.md
+++ b/content/spin/v2/javascript-components.md
@@ -17,12 +17,12 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/javascript
 - [Storing Data in SQLite](#storing-data-in-sqlite)
 - [Storing Data in MySQL and PostgreSQL Relational Databases](#storing-data-in-mysql-and-postgresql-relational-databases)
 - [AI Inferencing From JS/TS Components](#ai-inferencing-from-jsts-components)
-- [`Node.js` Compatibility](#nodejs-compatibility)
+- [Node.js Compatibility](#nodejs-compatibility)
 - [Using External NPM Libraries](#using-external-npm-libraries)
   - [Suggested Libraries for Common Tasks](#suggested-libraries-for-common-tasks)
 - [Caveats](#caveats)
 
-With JavaScript being a very popular language, Spin provides an SDK to support building components. The development of the JavaScript SDK is continually being worked on to improve user experience and add features. The SDK is based on [ComponentizeJS](https://github.com/bytecodealliance/ComponentizeJS).
+With JavaScript being a very popular language, Spin provides an SDK to support building components. The development of the JavaScript SDK is continually being worked on to improve user experience and add features. The SDK is based on [`ComponentizeJS`](https://github.com/bytecodealliance/ComponentizeJS).
 
 > This guide assumes you have Spin installed. If this is your first encounter with Spin, please see the [Quick Start](quickstart), which includes information about installing Spin with the JavaScript templates and creating JavaScript and TypeScript applications.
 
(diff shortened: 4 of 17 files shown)


content/spin/v2/build.md
```diff
@@ -66,23 +66,17 @@ command = "cargo build --target wasm32-wasi --release"
 
 {{ startTab "TypeScript" }}
 
-For JavaScript and TypeScript applications, you must have the `js2wasm` Spin plugin installed:
+For JavaScript and TypeScript applications, you must have `node.js`.
```
radu-matei, line 69: nit: Node.js (ref https://nodejs.org/en)

content/spin/v2/build.md
```diff
@@ -66,23 +66,17 @@ command = "cargo build --target wasm32-wasi --release"
...
 
-<!-- @selectiveCpy -->
-
-```bash
-$ spin plugins update
-$ spin plugins install js2wasm --yes
-```
 
-It's normally convenient to put the detailed build instructions in `package.json`:
+It's normally convenient to put the detailed build instructions in `package.json`, The build script calls out to [`webpack`](https://webpack.js.org/) and `j2w` which is a script provided by the `@fermyon/spin-sdk` package that utilizes [`ComponentizeJS`](https://github.com/bytecodealliance/ComponentizeJS). The build script looks like:
```
radu-matei, line 72: after `package.json` there is a coma that I think should be a full stop.

content/spin/v2/build.md
```diff
@@ -66,23 +66,17 @@ command = "cargo build --target wasm32-wasi --release"
...
 
 <!-- @nocpy -->
 
 ```json
 {
   "scripts": {
-    "build": "npx webpack --mode=production && mkdir -p target && spin js2wasm -o target/spin-http-js.wasm dist/spin.js"
+   "build": "npx webpack --mode=production && npx mkdirp target && npx j2w -i dist.js -d combined-wit -n combined -o target/spin-http-js.wasm"
```
radu-matei, line 79: Not a docs question, but I assume we're using `npx mkdirp` because it's cross platform?

content/spin/v2/http-outbound.md
```diff
@@ -102,9 +102,10 @@ const response = await fetch("https://example.com/users");
 
 **Notes**
 
-* Although the underlying Spin interface is blocking, the `fetch` function is defined by JavaScript as async. You must await the response, but the request will always block, and the promise will resolve as soon as the request is returned.
+// TODO: Add example and update link
```
radu-matei, line 105: TODO

content/spin/v2/javascript-components.md
```diff
@@ -18,13 +17,14 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/javascript
 - [Storing Data in SQLite](#storing-data-in-sqlite)
 - [Storing Data in MySQL and PostgreSQL Relational Databases](#storing-data-in-mysql-and-postgresql-relational-databases)
 - [AI Inferencing From JS/TS Components](#ai-inferencing-from-jsts-components)
+- [`Node.js` Compatibility](#nodejs-compatibility)
 - [Using External NPM Libraries](#using-external-npm-libraries)
   - [Suggested Libraries for Common Tasks](#suggested-libraries-for-common-tasks)
 - [Caveats](#caveats)
 
-With JavaScript being a very popular language, Spin provides support for building components with it using the experimental SDK. The development of the JavaScript SDK is continually being worked on to improve user experience and add features. 
+With JavaScript being a very popular language, Spin provides an SDK to support building components. The development of the JavaScript SDK is continually being worked on to improve user experience and add features. The SDK is based on [ComponentizeJS](https://github.com/bytecodealliance/ComponentizeJS).
```
radu-matei, line 25: In the previous section, it was `ComponentizeJS` in backticks. We should be consistent.

content/spin/v2/javascript-components.md
```diff
@@ -86,16 +77,17 @@ This creates a directory of the following structure:
...
 ├── src
-│   └── index.ts
+│   ├── index.ts
+│   └── spin.ts
 ├── tsconfig.json
 └── webpack.config.js
 ```
 
-The source for the component is present in `src/index.ts`. [Webpack](https://webpack.js.org) is used to bundle the component into a single `.js` file which will then be compiled to a `.wasm` module using the `js2wasm` plugin.
+The source for the component is present in `src/index.ts`. [Webpack](https://webpack.js.org) is used to bundle the component into a single `.js` file which will then be compiled to a `.wasm` module using the `j2w` node executable provided by the `@fermyon/spin-sdk` which is a wrapper around `ComponentizeJS`. The `knitwit.json` is the configuration file used by [knitwit](https://github.com/fermyon/knitwit) to manage the WebAssembly dependencies of each package.
```
radu-matei, line 90: Given the existing structure, we should treat this comment separately. But I'm not sure about including this kind of detail in the "installing templates" and "building your first app" section.

content/spin/v2/javascript-components.md
```diff
@@ -385,12 +326,45 @@ For more information about using relational databases from TypeScript/JavaScript
 
 For more information about using Serverless AI from JS/TS, see the [Serverless AI](serverless-ai-api-guide) API guide.
 
+## `Node.js` Compatibility
```
radu-matei, line 329: nit: not in backticks.

content/spin/v2/javascript-components.md
```diff
@@ -385,12 +326,45 @@ For more information about using relational databases from TypeScript/JavaScript
...
+
+The SDK does not support the full specification of `Node.js`. A limited set of APIs can be polyfilled using the [`@fermyon/wasi-ext`](https://github.com/fermyon/js-wasi-ext) library which provided a webpack plugin. It can be used by installing the  library first using:
```
radu-matei, line 331: "library which provided a webpack plugin"? Why the past tense?

radu-matei (commented):
Thanks for tackling this!
Left a few comments, some of which are not blocking.

content/spin/v2/build.md
karthik2804: Yep - We were originally using `mkdir -p` on the old SDK but that led to it breaking on Windows as it does not support the `-p` option.

content/spin/v2/javascript-components.md
karthik2804: typo - Should have been provides. will fix.

content/spin/v2/javascript-components.md
karthik2804: Happy to remove that but not sure if it would be better to create a new heading explaining how the different pieces fit together.

karthik2804:
@macolso That page does not have anything specific to JS/TS. It is only about the features Cloud supports.

content/spin/v2/http-trigger.md
```diff
@@ -267,31 +267,14 @@ For a full Rust SDK reference, see the [Rust Spin SDK documentation](https://doc
 
 > [**Want to go straight to the reference documentation?**  Find it here.](https://fermyon.github.io/spin-js-sdk/)
 
-In JavaScript or TypeScript, the handler is identified by name.  It must be called `handleRequest`.  The way you declare it is slightly different between the two languages.
+In TypeScript, the user must a define a function named `handler` which the SDK attaches to the [`fetch` event listener](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent).
 
-In **JavaScript**, `handleRequest` is declared as `export async function`.  It takes a JavaScript [object representing the request](https://fermyon.github.io/spin-js-sdk/interfaces/HttpRequest.html), and returns a [response object](https://fermyon.github.io/spin-js-sdk/interfaces/HttpResponse.html).  The fields of these objects are exactly the same as in TypeScript:
```
itowlson, line 272: I see you've removed the unwieldy wording about the types/shapes/APIs of the arguments (and the `export async` etc.).  Do you think we still need to say anything about these?

content/spin/v2/key-value-store-tutorial.md
```diff
@@ -76,7 +76,7 @@ $ spin new -t http-rust spin-key-value
 ```bash
 $ spin new -t http-ts spin-key-value
 
-# Reference: https://github.com/fermyon/spin-js-sdk/tree/main/examples/typescript/spin_kv
+# Reference: https://github.com/fermyon/spin-js-sdk/tree/feat/sdk-v2/examples/spin-host-apis/spin-kv
```
itowlson, line 79: Should this be `main` now?

content/spin/v2/key-value-store-tutorial.md
```diff
@@ -233,52 +233,49 @@ fn handle_request(req: Request) -> anyhow::Result<impl IntoResponse> {
...
       break;
     case "HEAD":
-      if (!store.exists(request.uri)) {
-        console.log(`Key ${request.uri} not found`);
+      if (!store.exists(req.uri)) {
+        console.log(`Key ${req.uri} not found`);
         status = 404
       } else {
-      console.log(`Found Key ${request.uri}`);
+      console.log(`Found Key ${req.uri}`);
```
itowlson, line 272: I know this was existing but should this line be indented?

content/spin/v2/kv-store-api-guide.md
```diff
@@ -85,38 +85,32 @@ fn handle_request(_req: Request) -> Result<impl IntoResponse> {
 
 > [**Want to go straight to the reference documentation?**  Find it here.](https://fermyon.github.io/spin-js-sdk/)
 
-With Typescript, the key value functions can be accessed after opening a store using either [the `Kv.open` or the `Kv.openDefault` methods](https://fermyon.github.io/spin-js-sdk/variables/Kv.html) which returns a [handle to the store](https://fermyon.github.io/spin-js-sdk/interfaces/_internal_.KvStore.html). For example:
+With Typescript, the key value functions can be accessed after opening a store using either [the `Kv.open` or the `Kv.openDefault` methods](https://fermyon.github.io/spin-js-sdk/modules/Kv.html) which returns a [handle to the store](https://fermyon.github.io/spin-js-sdk/interfaces/Kv.Store.html). For example:
```
itowlson, line 88: This looks like a residual "With TypeScript" for which there was no corresponding "With JavaScript" (and didn't need to be).  I'd remove it.

content/spin/v2/kv-store-api-guide.md
```diff
@@ -85,38 +85,32 @@ fn handle_request(_req: Request) -> Result<impl IntoResponse> {
...
 
-```javascript
-import { HandleRequest, HttpRequest, HttpResponse, Kv } from "@fermyon/spin-sdk"
+```ts
+import { ResponseBuilder } from "@fermyon/spin-sdk";
```
itowlson, line 91: Looking at the repo example shouldn't it be importing `Kv` as well here?

content/spin/v2/key-value-store-tutorial.md
```diff
@@ -233,52 +233,49 @@ fn handle_request(req: Request) -> anyhow::Result<impl IntoResponse> {
 {{ startTab "TypeScript"}}
 
 ```typescript
-import { HandleRequest, HttpRequest, HttpResponse, Kv } from "@fermyon/spin-sdk"
+import { ResponseBuilder } from "@fermyon/spin-sdk";
```
itowlson, line 236: No need to import `Kv` any more?

content/spin/v2/kv-store-api-guide.md
```diff
@@ -85,38 +85,32 @@ fn handle_request(_req: Request) -> Result<impl IntoResponse> {
...
     let store = Kv.openDefault()
     store.set("mykey", "myvalue")
-    return {
-            status: 200,
-            headers: {"content-type":"text/plain"},
-            body: store.get("mykey") ?? encoder.encode("Key not found")
-    }
+    res.status(200)
+    res.set({"content-type":"text/plain"})
+    res.send(store.get("mykey") ?? "Key not found")
```
itowlson, line 98: Previously this had an `encoder.encode` around the `"Key not found"` because bytes vs strings.  Is that no longer needed?

content/spin/v2/kv-store-api-guide.md
```diff
@@ -85,38 +85,32 @@ fn handle_request(_req: Request) -> Result<impl IntoResponse> {
...
 ```
 
 **General Notes**
-- The spinSdk object is always available at runtime. Code checking and completion are available in TypeScript at design time if the module imports anything from the @fermyon/spin-sdk package. For example: 
-- The JavaScript SDK doesn't surface the `close` operation. It automatically closes all stores at the end of the request; there's no way to close them early.
+- The SDK doesn't surface the `close` operation. It automatically closes all stores at the end of the request; there's no way to close them early.
 
-[`get` **Operation**](https://fermyon.github.io/spin-js-sdk/interfaces/_internal_.KvStore.html#get)
+[`get` **Operation**](https://fermyon.github.io/spin-js-sdk/interfaces/Kv.Store.html#get)
 - The result is of the type `ArrayBuffer | null`
```
itowlson, line 106: Now `Uint8Array | null`?

content/spin/v2/http-trigger.md
```diff
@@ -267,31 +267,14 @@ For a full Rust SDK reference, see the [Rust Spin SDK documentation](https://doc
...
-```javascript
-import { HandleRequest, HttpRequest, HttpResponse} from "@fermyon/spin-sdk"
-
-export const handleRequest: HandleRequest = async function(request: HttpRequest): Promise<HttpResponse> {
-    return {
-      status: 200,
-        headers: { "content-type": "text/plain" },
-      body: "Hello from TS-SDK"
-    }
+export async function handler(req: Request, res: ResponseBuilder) {
```
itowlson, line 275: `Request` is some kind of builtin, right?

content/spin/v2/rdbms-storage.md
```diff
@@ -90,22 +88,13 @@ const DB_URL = "mysql://root:@127.0.0.1/spin_dev"
...
-export const handleRequest: HandleRequest = async function (request: HttpRequest): Promise<HttpResponse> {
-
-    Mysql.execute(DB_URL, "delete from test where id=?", [4])
-    Mysql.execute(DB_URL, "insert into test values (4,5)", [])
-    let test = Mysql.query(DB_URL, "select * from test", [])
+export async function handler(_req: Request, res: ResponseBuilder) {
+  let conn = Mysql.open(DB_URL);
+  conn.execute('delete from test where id=?', [4]);
+  conn.execute('insert into test values (4,5)', []);
+  let ret = conn.query('select * from test', []);
```
itowlson, line 95: A comment showing what this looks like (or code showing how to use it) would be very useful here.  Looking at the help I _think_ there's an array of rows, each of which is an object keyed by column name (a la JSON `{ "columns": ???, "rows": [{ "id": 4, "val": 5 }] }` - is that right?

(I am not sure about the columns.  The reference suggests each column has a name and an _array_ of data types.  But I think I must be reading that wrong.)

content/spin/v2/redis-outbound.md
```diff
@@ -90,25 +91,23 @@ You can find a complete Rust code example for using outbound Redis from an HTTP
...
-Redis functions are available on [the `Redis` object](https://fermyon.github.io/spin-js-sdk/variables/Redis.html). The function names match the operations above, but you must pass the Redis instance address to _each_ operation as its first parameter. For example:
+Redis functions are available on [the `Redis` module](https://fermyon.github.io/spin-js-sdk/modules/Redis.html). The function names match the operations above. For example:
 
 ```javascript
-import {Redis} from "@fermyon/spin-sdk"
+import { Redis } from "@fermyon/spin-sdk"
 
-const value = Redis.get(address, key);
+let db = Redis.open("redis://localhost:6379")
+const value = db.get(key);
```
itowlson, line 100: Is there a reason for shifting between `let` and `const` here?

content/spin/v2/redis-outbound.md
```diff
@@ -90,25 +91,23 @@ You can find a complete Rust code example for using outbound Redis from an HTTP
...
 ```
 
 **General Notes**
-
-* The `spinSdk` object is always available at runtime. Code checking and completion are available in TypeScript at design time if the module imports anything from the `@fermyon/spin-sdk` package.
-* Address and key parameters are strings.
-* Bytes parameters and return values are buffers (TypeScript `ArrayBuffer`).
+* key parameters are strings.
```
itowlson, line 104:
```suggestion
* Key parameters are strings.
```

content/spin/v2/redis-outbound.md
```diff
@@ -90,25 +91,23 @@ You can find a complete Rust code example for using outbound Redis from an HTTP
...
+* Bytes parameters and return values are buffers (TypeScript `Uint8Array`).
 * Lists are passed and returned as arrays.
-* If a Spin SDK function fails, it throws an [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error).
```
itowlson, line 107: You've removed the error info - is that intentional?  Is this sort of thing considered implicit in JS?  (Genuine question, I have no idea what is idiomatic in JS docs except for being half-as-- er anyway business before RANTY PLEASURE.)

content/spin/v2/redis-outbound.md
```diff
@@ -90,25 +91,23 @@ You can find a complete Rust code example for using outbound Redis from an HTTP
...
 
 **`execute` Operation**
 
-* The arguments and results can be either numbers or buffers. (In TypeScript they are union types, e.g. `BigInt | ArrayBuffer`.)
+* The arguments and results can be either numbers or buffers. (In TypeScript they are union types, e.g. `BigInt | Uint8Array`.)
```
itowlson, line 110: The help reference says they are `RedisParameter` which is a 404.

(Huh. And that execute returns `RedisResult` which is also a 404.  Do these work for you?  https://fermyon.github.io/spin-js-sdk/interfaces/Redis.RedisConnection.html#execute)

content/spin/v2/redis-outbound.md
```diff
@@ -90,25 +91,23 @@ You can find a complete Rust code example for using outbound Redis from an HTTP
...
 
 You can find a complete TypeScript example for using outbound Redis from an HTTP component in the [JavaScript SDK repository on GitHub](https://github.com/fermyon/spin-js-sdk/tree/main/examples/typescript/outbound_redis). Please also see this, related, [outbound Redis (using TypeScript) section](./javascript-components#storing-data-in-redis-from-jsts-components).
```
itowlson, line 112: Dead link to the outbound Redis example.

content/spin/v2/sqlite-api-guide.md
```diff
@@ -120,24 +120,24 @@ struct ToDo {
...
+}
 ```
 
 **General Notes**
-* The `spinSdk` object is always available at runtime. Code checking and completion are available in TypeScript at design time if the module imports anything from the `@fermyon/spin-sdk` package.
-* Parameters are JavaScript values (numbers, strings, byte arrays, or nulls). Spin infers the underlying SQL type.
-* The `execute` function returns an object with `rows` and `columns` properties. `columns` is an array of strings representing column names. `rows` is an array of rows, each of which is an array of JavaScript values (as above) in the same order as `columns`.
-* The `SqliteStore` object doesn't surface the `close` function.
-* If a Spin SDK function fails, it throws an [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error).
+* The `execute` function returns an object with `rows` and `columns` properties. `columns` is an array of strings representing column names. `rows` is an array of rows, each of which is an object containing Javascript Values keyed using the column names.
```
itowlson, line 139:
```suggestion
* The `execute` function returns an object with `rows` and `columns` properties. `columns` is an array of strings representing column names. `rows` is an array of rows, each of which is an object containing Javascript values keyed using the column names.
```

content/spin/v2/http-trigger.md
```diff
@@ -267,31 +267,14 @@ For a full Rust SDK reference, see the [Rust Spin SDK documentation](https://doc
 
 > [**Want to go straight to the reference documentation?**  Find it here.](https://fermyon.github.io/spin-js-sdk/)
 
-In JavaScript or TypeScript, the handler is identified by name.  It must be called `handleRequest`.  The way you declare it is slightly different between the two languages.
+In TypeScript, the user must a define a function named `handler` which the SDK attaches to the [`fetch` event listener](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent).
```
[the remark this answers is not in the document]
karthik2804: A wasi-http incoming event is translated in the engine to be a fetchEvent ([ref](https://developer.mozilla.org/en-US/docs/Web/API/FetchEvent))

content/spin/v2/kv-store-api-guide.md
karthik2804: nope, `res.send()` takes both bytes and string.

content/spin/v2/http-trigger.md
karthik2804: Yes Request is a builtin. [ref](https://developer.mozilla.org/en-US/docs/Web/API/Request)

content/spin/v2/redis-outbound.md
karthik2804: nope I will make it consistent.

content/spin/v2/redis-outbound.md
karthik2804: I can reproduce it. I will take a look at the docs generated.

content/spin/v2/http-trigger.md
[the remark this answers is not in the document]
itowlson: Maybe writing it as "the `FetchEvent` listener" rather than "the `fetch` event" would be clearer, but again I don't know what is idiomatic in JS.  For me it would still invite the question "which FetchEvent listener" (because I know it as the HTTP trigger) but that may be another idiom thing!

content/spin/v2/http-outbound.md
```diff
@@ -102,9 +102,9 @@ const response = await fetch("https://example.com/users");
 
 **Notes**
 
-* Although the underlying Spin interface is blocking, the `fetch` function is defined by JavaScript as async. You must await the response, but the request will always block, and the promise will resolve as soon as the request is returned.
+You can find a complete example of using outbound HTTP in the JavaScript SDK repository on [GitHub](https://github.com/fermyon/spin-js-sdk/tree/main/examples/common-patterns/outbound-http)
 
-You can find a complete example of using outbound HTTP in the JavaScript SDK repository on GitHub ([TypeScript](https://github.com/fermyon/spin-js-sdk/tree/main/examples/typescript/outbound_http), [JavaScript](https://github.com/fermyon/spin-js-sdk/tree/main/examples/javascript/outbound-http)).
+**Note**: `fetch` currently only works when building for the HTTP trigger.
```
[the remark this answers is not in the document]
karthik2804: Is there some action item that is needed here?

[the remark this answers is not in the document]
itowlson: No, sorry, I was just astonished.  I will move the comment to #yelling.

content/spin/v2/build.md
```diff
@@ -66,23 +66,17 @@ command = "cargo build --target wasm32-wasi --release"
...
 
-<!-- @selectiveCpy -->
-
-```bash
-$ spin plugins update
-$ spin plugins install js2wasm --yes
-```
 
-It's normally convenient to put the detailed build instructions in `package.json`:
+It's normally convenient to put the detailed build instructions in `package.json`. The build script calls out to [`webpack`](https://webpack.js.org/) and `j2w` which is a script provided by the `@fermyon/spin-sdk` package that utilizes [`ComponentizeJS`](https://github.com/bytecodealliance/ComponentizeJS). The build script looks like:
```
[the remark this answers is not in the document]
karthik2804: I will add it to a details tab.

content/spin/v2/http-trigger.md
karthik2804: added a part of it back.

content/spin/v2/rdbms-storage.md
karthik2804: That was a typo. I will fix it in the docs. It should just be the name and the type of the column.

karthik2804: Fixed the ref docs.

content/spin/v2/redis-outbound.md
karthik2804: fixed the ref docs.

content/spin/v2/http-outbound.md
```diff
@@ -102,9 +102,9 @@ const response = await fetch("https://example.com/users");
 
 **Notes**
 
-* Although the underlying Spin interface is blocking, the `fetch` function is defined by JavaScript as async. You must await the response, but the request will always block, and the promise will resolve as soon as the request is returned.
+You can find a complete example of using outbound HTTP in the JavaScript SDK repository on [GitHub](https://github.com/fermyon/spin-js-sdk/tree/main/examples/common-patterns/outbound-http)
```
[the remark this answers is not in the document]
karthik2804: The `spin.ts` only wraps the user provided handler and attaches it to the FetchEvent.

content/spin/v2/redis-outbound.md
karthik2804: It is normal for JS to throw errors when things fail. I was not sure adding that generic information was going to provide much value but happy to add it back.

content/spin/v2/http-outbound.md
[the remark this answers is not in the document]
itowlson: Oh!  So the "true" entry point is the `handleEvent` function, but that's unfriendly, so we shunt it off into this "pay no attention to the man behind the curtain" file, which hands off to the friendly entry point?  And we can't do this inside the SDK because it needs to import the user handler?

So this is what the stuff about "the SDK attaches `handle` to the FetchEvent" thing was - "the SDK" here is this user-side bit rather than the `@fermyon/spin-sdk` library!  This all makes way more sense now!

This is non-blocking but I think

1. we should explain this, at least in passing, in (a relevant part of) the docs
2. we should add a comment to the template file about it being required plumbing with which they should not mess

Or maybe this is all just obvious to JS folks, it just wasn't to me...!

content/spin/v2/redis-outbound.md
itowlson: Nah, no worries.

itowlson (approved):
Great stuff!  Thanks for working through all this.  The new SDK looks nice!

PR #1356, opened 2024-07-31
Update variables documentation with latest APIs and better example

kate-goldenring:
Our variables example showed a bad example of checking passwords -- it does not at minimum do a constant time comparison. While it is an example, we cannot assume that folks will not use it. Instead, this updates the example to be a more realistic one that calls an API with a token. 

This also updates the examples to use the latest SDKs -- Python was out of date -- and talks about templating within the manifest.

I've run all examples locally -- here is a repo with the examples for if we need to test them for future sdk changes: https://github.com/kate-goldenring/spin-variables-examples

- [ ] File does not use CRLF, but uses plain LF (hint: use `cat -ve <filename> | grep $'\r' | wc -l` and expect 0 as a result)

### DIFF
--- a/content/spin/v2/variables.md
+++ b/content/spin/v2/variables.md
@@ -11,7 +11,7 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/variables.
 
 Spin supports dynamic application variables. Instead of being static, their values can be updated without modifying the application, creating a simpler experience for rotating secrets, updating API endpoints, and more. 
 
-These variables are defined in a Spin application manifest (in the `[variables]` section), and their values can be set or overridden at runtime by an [application variables provider](./dynamic-configuration.md#application-variables-runtime-configuration). When running Spin locally, the variables provider can be [Hashicorp Vault](./dynamic-configuration.md#vault-application-variable-provider) for secrets, or host environment variables.
+These variables are defined in a Spin application manifest (in the `[variables]` section), and their values can be set or overridden at runtime by an [application variables provider](./dynamic-configuration.md#application-variables-runtime-configuration). When running Spin locally, the variables provider can be [Hashicorp Vault](./dynamic-configuration.md#vault-application-variable-provider) for secrets, [Azure Key Vault](https://azure.microsoft.com/en-us/products/key-vault), or host environment variables.
 
 For information about configuring application variables providers, refer to the [dynamic configuration documentation](./dynamic-configuration.md#application-variables-runtime-configuration).
 
@@ -19,7 +19,11 @@ For information about configuring application variables providers, refer to the 
 
 Variables are added to an application under the top-level `[variables]` section of an application manifest (`spin.toml`). Each entry must either have a default value or be marked as `required = true`. “Required” entries must be [provided](./dynamic-configuration#application-variables-runtime-configuration) with a value.
 
-For example, say an application needs to access a bearer token for an outbound API call. A `[variables]` section could be added to an application's manifest with one entry for a variable named `token`. Since there is no reasonable default value for a secret, the variable is set as required with `required = true`. On the other hand, say the the API endpoint is known but should be configurable for A/B testing, another v`api_uri` variable could be added with a default that can be overridden. This scenario could be configured in the application manifest as follows:
+For example, consider an application which makes an outbound API call using a bearer token. To configure this using variables, you would:
+* Add a `[variables]` section to the application manifest
+* Add a `token` entry for the bearer token.  Since there is no reasonable default value for this secret, set the variable as required with `required = true`.
+* Add an `api_uri` variable.  The URL _is_ known, but is useful to override, for example for A/B testing. So you can give this variable a default value with `default = "http://my-api.com"`.
+The resulting application manifest looks like this:
 
 <!-- @nocpy -->
 
@@ -40,7 +44,9 @@ api_uri = "\{{ api_uri }}"
 api_version = "v1"
 ```
 
-Variables can also be surfaced to other sections of the application manifest that benefit from dynamic configuration. For example, the `allowed_outbound_hosts` can be dynamically configured using variables as follows:
+When a component variable references an application variable, its value will dynamically update as the application variable changes. For example, if the `api_token` variable is provided using the [Spin Vault provider](./dynamic-configuration.md#vault-application-variable-provider), it can be updated by changing the value in HashiCorp Vault. The next time the component gets the value of `token`, the latest value of `api_token` will be returned by the provider. See the [next section](#using-variables-from-applications) to learn how to use Spin's configuration SDKs to get configuration variables within applications.
+
+Variables can also be used in other sections of the application manifest that benefit from dynamic configuration. In these cases, the variables are substituted at application load time rather than dynamically updated while the application is running. For example, the `allowed_outbound_hosts` can be dynamically configured using variables as follows:
 
 <!-- @nocpy -->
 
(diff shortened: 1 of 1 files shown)


content/spin/v2/variables.md
```diff
@@ -11,60 +11,72 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/variables.
 
 Spin supports dynamic application variables. Instead of being static, their values can be updated without modifying the application, creating a simpler experience for rotating secrets, updating API endpoints, and more. 
 
-These variables are defined in a Spin application manifest (in the `[variables]` section), and their values can be set or overridden at runtime by an [application variables provider](./dynamic-configuration.md#application-variables-runtime-configuration). When running Spin locally, the variables provider can be [Hashicorp Vault](./dynamic-configuration.md#vault-application-variable-provider) for secrets, [Azure Key Vault](https://azure.microsoft.com/en-us/products/key-vault), or host environment variables.
+These variables are defined in a Spin application manifest (in the `[variables]` section), and their values can be set or overridden at runtime by an [application variables provider](./dynamic-configuration.md#application-variables-runtime-configuration). When running Spin locally, the variables provider can be [Hashicorp Vault](./dynamic-configuration.md#vault-application-variable-provider) for secrets, or host environment variables.
```
itowlson, line 14: Azure KeyVault is still there on main... are you saying this is being removed?  Is that part of the factors work?  Can we discuss?

content/spin/v2/variables.md
```diff
@@ -11,60 +11,72 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/variables.
...
 
 Variables are added to an application under the top-level `[variables]` section of an application manifest (`spin.toml`). Each entry must either have a default value or be marked as `required = true`. “Required” entries must be [provided](./dynamic-configuration#application-variables-runtime-configuration) with a value.
 
-For example, say an application needs to access a secret. A `[variables]` section could be added to an application's manifest with one entry for a variable named `secret`. Since there is no reasonable default value for a secret, the variable is set as required with `required = true`:
+For example, say an application needs to access a bearer token for an outbound API call. A `[variables]` section could be added to an application's manifest with one entry for a variable named `token`. Since there is no reasonable default value for a secret, the variable is set as required with `required = true`. On the other hand, say the the API endpoint is known but should be configurable for A/B testing, another v`api_uri` variable could be added with a default that can be overridden. This scenario could be configured in the application manifest as follows:
```
itowlson, line 22:
```suggestion
For example, say an application needs to access a bearer token for an outbound API call. A `[variables]` section could be added to an application's manifest with one entry for a variable named `token`. Since there is no reasonable default value for a secret, the variable is set as required with `required = true`. On the other hand, say the the API endpoint is known but should be configurable for A/B testing, another `api_uri` variable could be added with a default that can be overridden. This scenario could be configured in the application manifest as follows:
```

itowlson, line 22: This feels like a lot of conditional mood ("could be added", "could be configured") - I wonder if we could make it read more positively and concretely.

```suggestion
For example, consider an application which makes an outbound API call using a bearer token. To configure this using variables, you would:

* Add a `[variables]` section to the application manifest
* Add a `token` entry for the bearer token.  Since there is no reasonable default value for this secret, set the variable as required with `required = true`.
* Add an `api_uri` variable.  The URL _is_ known, but is useful to override, for example for A/B testing. So you can give this variable a default value with `default = "http://my-api.com"`.

The resulting application manifest looks like this:
```

Maybe that is a bit _too_ definite though!

content/spin/v2/variables.md
```diff
@@ -11,60 +11,72 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/variables.
...
-password = "\{{ secret }}"
-api_host = "https://my-api.com"
+[component.(name).variables]
+token = "\{{ api_token }}"
+api_uri = "\{{ api_uri }}"
+api_version = "v1"
 ```
 
-When a component variable references an application variable, its value will dynamically update as the application variable changes. For example, if the `secret` variable is provided using the [Spin Vault provider](./dynamic-configuration.md#vault-application-variable-provider), it can be updated by changing the value in HashiCorp Vault. The next time the component gets the value of `password`, the latest value of `secret` will be returned by the provider. See the [next section](#using-variables-from-applications) to learn how to use Spin's configuration SDKs to get configuration variables within applications.
+Variables can also be surfaced to other sections of the application manifest that benefit from dynamic configuration. For example, the `allowed_outbound_hosts` can be dynamically configured using variables as follows:
```
itowlson, line 43:
```suggestion
Variables can also be used in other sections of the application manifest that benefit from dynamic configuration. For example, the `allowed_outbound_hosts` can be dynamically configured using variables as follows:
```

content/spin/v2/variables.md
```diff
@@ -11,60 +11,72 @@ url = "https://github.com/fermyon/developer/blob/main/content/spin/v2/variables.
...
 
-An application manifest with a `secret` variable and a component that uses it would look similar to the following:
+<!-- @nocpy -->
+
+```toml
+[component.(name)]
+allowed_outbound_hosts = [ "\{{ api_uri }}" ]
+```
+
+When a component variable references an application variable, its value will dynamically update as the application variable changes. For example, if the `api_token` variable is provided using the [Spin Vault provider](./dynamic-configuration.md#vault-application-variable-provider), it can be updated by changing the value in HashiCorp Vault. The next time the component gets the value of `token`, the latest value of `api_token` will be returned by the provider. See the [next section](#using-variables-from-applications) to learn how to use Spin's configuration SDKs to get configuration variables within applications.
```
itowlson, line 52: This refers specifically to component variables referencing application variables, so goes best either with the section on component variables (where it was before), or in a new section separated from the "variables in other parts of the manifest" stuff.

I think I'd prefer to move it back, and then we can add something here that variables in other parts of the manifest are substituted at application load time and not dynamically updated while the application is running.  Open to thoughts though!

content/spin/v2/variables.md
```diff
@@ -117,28 +128,31 @@ fn handle_spin_example(req: Request) -> Result<impl IntoResponse> {
...
 export const handleRequest: HandleRequest = async function (request: HttpRequest): Promise<HttpResponse> {
-  const expected = decoder.decode(request.body)
-  let password = Config.get("password")
-  let access = "denied"
-  if (expected === password) {
-      access = "accepted"
-  }
-  let responseJson = `{\"authentication\": \"${access}\"}`;
-
+  let token = Config.get("token")
```
itowlson, line 139: Heads up that this is changing in JS SDK v2 and there is #1355 in progress for that, so some synchronisation might be needed...

content/spin/v2/variables.md
```diff
@@ -149,22 +163,28 @@ export const handleRequest: HandleRequest = async function (request: HttpRequest
 
 > [**Want to go straight to the reference documentation?**  Find it here.](https://fermyon.github.io/spin-python-sdk/variables.html)
 
-The `variables` module has a function called `get`(https://fermyon.github.io/spin-python-sdk/variables.get).
+The `variables` module has a function called `get`(https://fermyon.github.io/spin-python-sdk/variables.html#spin_sdk.variables.get).
```
itowlson, line 166: Not your problem but I wonder why the link checker wasn't flagging the old broken link...

content/spin/v2/variables.md
```diff
@@ -215,22 +247,22 @@ To build and run the application, we issue the following commands:
 ```bash
 # Build the application
 $ spin build
-# Export the value of the secret, using the `SPIN_VARIABLE` prefix (upper case is necessary as shown here)
-$ export SPIN_VARIABLE_SECRET="your-password-here"
-# Run the application
-$ spin up
+# Run the application, setting the values of the API token and URI via the environment variable provider
+# using the `SPIN_VARIABLE` prefix (upper case is necessary as shown here)
+$ SPIN_VARIABLE_API_TOKEN="your-token-here" SPIN_VARIABLE_API_URI="http://my-best-api.com spin up
```
itowlson, line 252:
```suggestion
$ SPIN_VARIABLE_API_TOKEN="your-token-here" SPIN_VARIABLE_API_URI="http://my-best-api.com" spin up
```

itowlson (changes requested):
This is great!  Thank you for reworking this, the new example is so much better.  Just a couple of niggles I'm afraid.

itowlson:
I think all the CI failures are changes to the JS SDK reference which will get fixed via #1355

content/spin/v2/variables.md
kate-goldenring: I originally started editing the `v1` file by accident and then copied it over. I clearly skipped over this v2 change that existed -- thank you!

content/spin/v2/variables.md
kate-goldenring: I love it! You are a word artist.

content/spin/v2/variables.md
kate-goldenring: Good call. I agree that it fits better right after the component section.

content/spin/v2/variables.md
kate-goldenring: Rebased!

kate-goldenring:
@itowlson this should be ready for another review

itowlson (approved):

PR #1367, opened 2024-08-20
docs(spin/v2): update cargo install steps and version strings

vdice:
- Updates the cargo install steps to remove the outdated/hard-coded version tag
  - This is one approach to avoid needing to updating a hard-coded string... but then again there are others as well; so we could just decide to keep it around and add steps in the release process doc to update.
- Updates some older version strings

### DIFF
--- a/content/spin/v2/install.md
+++ b/content/spin/v2/install.md
@@ -66,12 +66,12 @@ $ sudo mv spin /usr/local/bin/
 
 > If you have already installed Spin by building from source, and then install it via the installer, we recommend you remove the older source install by running `cargo uninstall spin-cli`  Otherwise the Cargo path may take precedence over the "install from binary" path, and commands may get the "wrong" version of Spin. Use `spin --version` to confirm the version on the PATH is the one you intend, or `which spin` to confirm the path it is found on.
 
-To install a specific version, you can pass arguments to the install script this way:
+To install a specific version (`v1.2.3` is just an example), you can pass arguments to the install script this way:
 
 <!-- @selectiveCpy -->
 
 ```bash
-$ curl -fsSL https://developer.fermyon.com/downloads/install.sh | bash -s -- -v v1.5.0
+$ curl -fsSL https://developer.fermyon.com/downloads/install.sh | bash -s -- -v v1.2.3
 ```
 
 To install the canary version of spin, you should pass the argument `-v canary`. The canary version is always the latest commit to the main branch of Spin:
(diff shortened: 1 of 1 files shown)


karthik2804 (approved):

content/spin/v2/install.md
```diff
@@ -212,8 +212,10 @@ If you have [`cargo`](https://doc.rust-lang.org/cargo/getting-started/installati
 <!-- @selectiveCpy -->
 
 ```bash
-$ git clone https://github.com/fermyon/spin -b v2.0.0
+$ git clone https://github.com/fermyon/spin
 $ cd spin
+$ # check out the latest tagged release
+$ git checkout $(git tag --sort=taggerdate | tail -1)
```
itowlson, line 218: I don't think this will work on Windows.  I guess we could just put a comment saying "on Windows, find the latest tag from the GH releases page"?  Hard to show both without making selective copy mad though.  I guess we could use a tab block but it all starts to get a bit dreary then...!

karthik2804: I think you can start a line with# and selective copy should be happy iirc.

vdice: Updated with revised wording.  What do you all think?

As a follow-up, we could add automation so that on any Spin release, a PR is triggered in this repo to bump hard-coded versions in this doc (or any docs).  We could then return to `git clone https://github.com/fermyon/spin -b <tag>`.  Overkill?

content/spin/v2/install.md
```diff
@@ -135,7 +135,7 @@ To install a specific version, you can pass arguments to the install script this
 <!-- @selectiveCpy -->
 
 ```bash
-$ curl -fsSL https://developer.fermyon.com/downloads/install.sh | bash -s -- -v v1.5.0
+$ curl -fsSL https://developer.fermyon.com/downloads/install.sh | bash -s -- -v v2.7.0
```
itowlson, line 138: Oh wait... I just noticed the context for this is "to install a specific version".  There's no need to keep it revved to latest version because that's the one they get by default.  I wonder if it would be better to make this an obviously (cough) fake version like 99.99.99 or 12.34.5?

content/spin/v2/install.md
```diff
@@ -207,13 +207,17 @@ This does not install any Spin templates or plugins. For a starter list, see the
...
+
+> Note: The `main` branch may have unstable or breaking changes. It is advised to check out the latest Spin release tag, which can be found by navigating to https://github.com/fermyon/spin/releases/latest.
 
 <!-- @selectiveCpy -->
 
 ```bash
-$ git clone https://github.com/fermyon/spin -b v2.0.0
+$ git clone https://github.com/fermyon/spin
 $ cd spin
+$ # Check out the latest tagged release (Linux/Mac)
```
itowlson, line 219: This still leaves Windows users with `main` though?

Given the aside you add above, I wonder if changing line 217 to use the fake version number (maybe with a comment like "change this to the latest tag as shown above" or something) is going to end up simpler?

Although I am probably overthinking this.  Such bike, very shed.

vdice: Ok, updated.  Please take a look when convenient.

itowlson (approved):
Thanks for your patience @vdice.  There's no really good solution here and I realise I've given sometimes contradictory feedback.  Let's get this in and see how we go.

PR #1395, opened 2024-10-08
fix if else clause in spin_main.hbs

karthik2804:
@itowlson here are the crimes needed to make the spin sidebar work with v3.


itowlson (approved):
CRIMETASTIC

Would you be able to do this in `templates/page_lang.hbs` please, or would you prefer me to apply the same fix there directly?  Thanks either way!

itowlson:
I've applied the fix to `page_lang.hbs` on the parent branch.  Thanks again for figuring this one out!

PR #1403, opened 2024-10-30
docs(wasm-languages): go updates

vdice:
This PR originally started with the goal of updating the example manifest to v2 and removing the `base` trigger config due to be removed in Spin 3 (ref https://github.com/fermyon/developer/issues/1376#issuecomment-2433510305).

I coudn't help but update a few tangential items along the way, but stopped short of revising the go/tinygo lay-of-the-land.

We can either collaborate in this PR to update the main text to reflect recent go/tinygo updates/support (pinging @adamreese @rajatjindal and others) or this can be a follow-up.

PR #1406, opened 2024-10-30
docs(wasm-languages): ruby updates

vdice:
Sample setup/code/manifest updates similar to https://github.com/fermyon/developer/pull/1403

Bonus: I did see the following error in the app logs when pinging the endpoint.  Cause for alarm?  Fixable?  (Response was still a 200 and contained the expected "Hello, Fermyon" body)

```
$ spin up

Serving http://127.0.0.1:3000
Available Routes:
  ruby: http://127.0.0.1:3000
2024-10-30T22:37:06.498213Z ERROR spin_http::wagi: Invalid header name error=invalid HTTP header name header_name=ruby 3.4.0dev (2024-06-28T17
```

### DIFF
--- a/content/wasm-languages/ruby.md
+++ b/content/wasm-languages/ruby.md
@@ -79,7 +79,7 @@ $ cd hello-ruby
 ```
 
 Now fetch a copy of the Ruby source from the [official releases](https://github.com/ruby/ruby.wasm/releases).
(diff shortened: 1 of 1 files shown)


itowlson (approved):

itowlson:
Could you raise an issue on Spin for the header error please?  It seems like it's just a quirk, but at least then we'll have a reminder to check it out.  Thanks!

PR #1424, opened 2024-11-25
fix: question mark for placeholder does not work

vanvuongngo:
# fix: question mark for placeholder does not work

If I follow the example with a question mark for a placeholder then this exception is thrown.

```ts
connection.execute("delete from tb_test where id=?", ['test3']);
```

```ts
"[object Object]" is not one of the cases of error
```

or this with a try-catch

error:
```ts
(new Error("[object Object] (see error.payload)", "dist.bindings.js", 157))
```
error.payload:
```ts
{ tag: "query-failed", val: "Error { kind: Db, cause: Some(DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E42601), message: "syntax error at end of input", detail: None, hint: None, position: Some(Original(35)), where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("scan.l"), line: Some(1240), routine: Some("scanner_yyerror") }) }" }
```

## environment

- spin: 3.0.0
- macos
- postgresql 17
- target: wasm32-wasip1

### DIFF
--- a/content/spin/v3/rdbms-storage.md
+++ b/content/spin/v3/rdbms-storage.md
@@ -99,7 +99,7 @@ export async function handler(_req: Request, res: ResponseBuilder) {
     res.send(JSON.stringify(ret, null, 2));
   } catch (error) {
     console.log(error.payload);
-	res.send({});
+	res.send('{}');
   }
 }
 ```
--- a/content/spin/v3/rdbms-storage.md
+++ b/content/spin/v3/rdbms-storage.md
@@ -89,13 +89,18 @@ const DB_URL = "mysql://root:@127.0.0.1/spin_dev"
 */
 
 export async function handler(_req: Request, res: ResponseBuilder) {
-  let conn = Mysql.open(DB_URL);
-  conn.execute('delete from test where id=$1', [4]);
-  conn.execute('insert into test values (4,5)', []);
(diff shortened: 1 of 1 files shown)


vanvuongngo (commented):
add try-catch to get error informations

itowlson:
Thanks for flagging this!  However, the page in question covers both MySQL and Postgres.  The example in this PR is for MySQL, for which the question mark is correct.  And we didn't want to clutter the page with MySQL and Postgres examples that were almost identical, so instead the paragraph above the sample links to a Postgres example.

I'll look at adding comments to the sample noting differences for Postgres.

@karthik2804 is it worth surfacing the error payload more visibly in this case, or is that part of the JS exception system and out of our control?

karthik2804:
@itowlson That error is the error generated as part of the bindings in `ComponentizeJS`. I think we can use try catch to make it clearer maybe but will have to take a look.

vanvuongngo:
Ah good catch and I did not realize that it was for MySql only. It took me awhile to figure out a solution w/o any documentation for postgresql.

itowlson:
@vanvuongngo No worries.  I'm proposing #1426 as a fix instead if you want to weigh in on that.  And we appreciate you flagging that this was confusing!

PR #1454, opened 2025-01-21
Fix search to only show releavnt v2 or v3 Spin docs

### LINKED ISSUE #1453: Fix search to only show result from current Spin version
For developer docs we wanted to only show results for Spin v1 if we were actively looking at Spin v1 docs, We need to add the same switch for v2 there.

https://github.com/fermyon/developer/blob/be747873e0ea6d4a54fe74df77ff8bb58dcba479/static/js/src/modules/search.js#L23-L34

mikkelhegn:
Fixes #1453

### DIFF
--- a/static/js/src/modules/search.js
+++ b/static/js/src/modules/search.js
@@ -1,301 +1,402 @@
-const { el, mount, text, list, setChildren, setStyle, setAttr } = redom
+const { el, mount, text, list, setChildren, setStyle, setAttr } = redom;
 
-const projectList = ["Spin", "Cloud", "Bartholomew"]
+const projectList = ["Spin", "Cloud", "Bartholomew"];
 const versionPattern = /^v\d+$/;
-let idx
-let documents
+let idx;
+let documents;
 
 async function getSearchIndex() {
-    try {
-        let res = await fetch('/static/data.json')
-        return await res.json()
-    }
-    catch (err) {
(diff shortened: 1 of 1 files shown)


static/js/src/modules/search.js
```diff
@@ -1,301 +1,402 @@
...
+      return k.url.includes("spin/v3/");
+    });
+  } else {
+    documents = documents.filter((k) => {
+      return (
+        k.project != "spin" ||
+        !k.url.includes("spin/v1/") ||
+        !k.url.includes("spin/v2/") ||
+        !k.url.includes("spin/v3/")
+      );
```
karthik2804, line 50: In other case where not in Spin docs, I think we want to include only v3 docs. 
```suggestion
    documents = documents.filter((k) => {
      return (
        k.project != "spin" ||
        k.url.includes("spin/v3/")
      );
```

karthik2804 (changes requested):
One tiny fix needed, everything else looks good.

karthik2804 (approved):

mikkelhegn:
Broken link in build is ok when I try this from here. Am I good to go ahead and merge, or should I rerun the build? @karthik2804

karthik2804:
It is okay to merge, I will add a follow up PR that adds that false positive link to the exclude list.

PR #1488, opened 2025-03-27
Add App Size limit

mikkelhegn:
According to this page, this is also a limit: https://www.fermyon.com/wasm-functions


mikkelhegn:
What's up with the linting errors? Should I be concerned?

flynnduism:
lgtm

linting not a blocker iMO

flynnduism (approved):

itowlson:
@mikkelhegn for some reason the linter has suddenly run berserk, apparently without us changing anything.  For the Spin docs we've disabled it for now until we can figure out why it's mad.

Issue #1490, opened 2025-03-27
spin.toml: update spin-related redirects


vdice:
Update spin-related redirects in spin.toml to point to https://spinframework.dev

PR #1497, opened 2025-03-27
ref(spin.toml): update with spinframework redirects

vdice:
- Updates Spin docs redirects to point to spinframework.dev
- Adds a catch-all `/spin/...` redirector after the more-specific ones preceeding it... _see further discussion [below](https://github.com/fermyon/developer/pull/1497#issuecomment-2773905474)_
- Removes triggers/components now handled by the new Spin docs site at https://github.com/spinframework/spin-docs/blob/main/spin.toml

Needs:
- [x] https://github.com/spinframework/spin-docs/issues/11
- [x] https://github.com/fermyon/spin-redirect/pull/12

Closes https://github.com/fermyon/developer/issues/1490


vdice:
Opening up for review, but again needs https://github.com/spinframework/spin-docs/issues/11 before merging.

adamreese (approved):
Manually checked

vdice:
(pushed rebase commit to get broken links fixes; no other changes)

vdice:
I do want to mention that the,catch-all `/spin/...` redirector isn't smart... it will always redirect to spinframework.dev, not persisting whatever path happened to be provided.  We'd need to look into updating the spin-redirect module to support [wildcard redirects](https://github.com/fermyon/spin-redirect/issues/4) -- but even then, we'd also need something on the spin-docs side to redirect any inbound `spin/...` request to `/...` (i.e. removing the `spin/` path prefix).

This _should_ be okay since there _shouldn't_ be any local `/spin` references after https://github.com/fermyon/developer/pull/1503 (and https://github.com/fermyon/developer/pull/1513).

itowlson:
> there shouldn't be any local /spin references

There will likely be _external_ /spin references, though, on sites we don't control (or don't want to spend ages updating).  Will they redirect correctly?

vdice:
>  There will likely be external /spin references, though, on sites we don't control (or don't want to spend ages updating). Will they redirect correctly?

Ah, yes, good reminder Ivan.

Another option came to mind: we could add redirects for each of the single-path, top-level sections (eg `spin/foo`, `spin/bar`, etc) to their appropriate destinations at spinframework.dev.

Or, we could look into adding wildcard support to the spin-redirect module (again, I believe it would need a corresponding redirect on the spin-docs side to remove the `spin/` path prefix).

WDYT?

itowlson:
We already kerjiggered the redirect module on the `spinframework.dev` side, we can totally kerjigger it here!

I think we'd looking at something like `/spin/...` -> `sf.dev/${wildcard)` which doesn't sound unreasonably onerous?  That seems like it would avoid the need for a further redirect on the destination side when the path is already versioned (and the destination already handles redirects if the path is unversioned).  Does that sound right?

vdice:
Added a note above re: needing finalized wildcard redirect functionality (ref https://github.com/fermyon/spin-redirect/pull/12).  Once that is merged, I'll update the wasm module to the one built from main of that project.

PR #1503, opened 2025-03-28
docs(*): update spin refs to spinframework.dev

vdice:
Closes https://github.com/fermyon/developer/issues/1494

- Updates Spin references to point to spinframework.dev
- ~~Note: removed some pages that themselves relied on relative Spin doc content, such as `content/{cloud|bartholomew|wasm-function}/contributing-docs.md`.  If we wish for another shared Contributing Docs page, we'll need to add one as a follow-up~~ _Update: I moved the existing contributing-docs.md file to /content/common and updated appropriate files._

Besides a few minor template updates, didn't attempt the full overhaul of site layout/templates/search functionality; tracking this in https://github.com/fermyon/developer/issues/1504

Depends on:
- [x] https://github.com/spinframework/spin-docs/issues/11
- [x] https://github.com/fermyon/developer/pull/1497

PR #1521, opened 2025-04-10
fix: fermyon/spin -> spinframework/spin

bacongobbler:
All of the repositories changed in this commit have been migrated to the spinframework org.

This coincides with other changes like #1503.

### DIFF
--- a/content/api/hub/contributing.md
+++ b/content/api/hub/contributing.md
@@ -140,7 +140,7 @@ The following is a list of the types of content that a user can submit to the Sp
     - Spin Application examples
     - Code snippets
 - **(Spin) Templates**
-    - Complete components such as the [Spin static-fileserver](https://github.com/fermyon/spin-fileserver)
+    - Complete components such as the [Spin static-fileserver](https://github.com/spinframework/spin-fileserver)
 - **(Spin) Plugins**
     - Triggers and tools
     - See Lee-Orr’s Message trigger: [https://github.com/lee-orr/spin-message-trigger](https://github.com/lee-orr/spin-message-trigger)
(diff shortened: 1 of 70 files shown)


bacongobbler:
CI will go green once https://github.com/spinframework/spin-go-sdk/pull/21 has been merged.

vdice:
Thanks for these updates @bacongobbler.

Question around the updates involving the spin go sdk:  With the search/replace approach, URLs are updated to eg `github.com/spinframework/spin/sdk/go/v2/http`; yet the go sdk has since been broken out to https://github.com/spinframework/spin-go-sdk.  Should we update appropriately?  eg `spinframework/spin/sdk/go/v2` -> `spinframework/spin-go-sdk`, etc?

bacongobbler:
Good catch. Yes, the new module name is `github.com/spinframework/spin-go-sdk/http` so the import paths should be updated to reflect that.

https://github.com/spinframework/spin-go-sdk/blob/f911142cc434d007c25eb03d11057c92a7f322d7/examples/http-outbound/http-to-same-app/main.go#L7

vdice:
> Good catch. Yes, the new module name is github.com/spinframework/spin-go-sdk/http so the import paths should be updated to reflect that.

Actually, as I was revising import paths, I realized I shouldn't touch the docs in the older `spin/v1` and `spin/v2` folders, since the examples/usage may be particular to the go sdk versions used at the time.  So, assuming the older go sdk versions/URLs can still be used (i.e. they aren't broken), we should leave these be.

Then, while updating the current `spin/v3` docs, I realized the go templates in the spin repo still reference the v2 version of the go sdk (see also https://github.com/spinframework/spin/issues/3106).  That, combined with a current unfamiliarity as to whether examples/usage will need to change when bumping to the [new go sdk](https://github.com/spinframework/spin-go-sdk) and the new sdk doesn't yet appear to have a tag or versioned release, I figured we should leave these be until 1. the go templates are updated and 2. we're ready for a first versioned release of the new go sdk (also, perhaps we're waiting for [wasip2 support](https://github.com/spinframework/spin-go-sdk/pulls?q=is%3Apr+is%3Aopen+wasip2) to land)  cc @adamreese

Mind you, we're also about to remove all `spin/*` directories in https://github.com/fermyon/developer/pull/1512 😉 so mostly the above applies to the spin documentation in its new home: https://github.com/spinframework/spin-docs

Therefore, I've only rebased this branch to fix the conflicts on main and will merge in its current form.  Thanks again!

vdice (approved):

Issue #1493, opened 2025-03-27
Update Spin Up Hub UI assets/links


vdice:
Update the Spin Up Hub UI assets/links to navigate to its new home at spinframework.dev

Depends on

- [x] https://github.com/spinframework/spin-docs/issues/11
- [x] https://github.com/spinframework/spin-docs/pull/47
- [x] https://github.com/spinframework/spin-docs/issues/91

vdice:
Noting that we'll potentially want to update the look of the cards in this section to indicate that clicking them will take the user to an external website (https://spinframework.dev/hub...)

<img width="1075" alt="Image" src="https://github.com/user-attachments/assets/a99e5d08-24cd-4106-a6f9-de2bbed77eac" />

PR #1539, opened 2025-04-28
chore(*): remove spin-up-hub content; add redirects; update URLs, etc

vdice:
- Removes `content/api/hub`, `spin-up-hub` directories
- Adds a wildcard `/hub/...` redirect to `spinframework.dev/hub/...`
- Updates URLs


vdice:
This is ready for review, though not sure if we want https://github.com/fermyon/developer/issues/1493 as a prerequisite.  Without 1493, the 'Spin Hub' button at top of site and corresponding cards in the body will whisk the user to https://spinframework.dev/hub (won't have the UI element indicating they are leaving the page).

karthik2804 (approved):
LGTM Thanks!

PR #1552, opened 2025-06-06
Bring the fwf quickstart install script into repo

karthik2804:
This commit brings in the install script into the repo and also updates it to install the plugin from the registry so that users will be "badgered" by the spin plugins system to update when there is a new version.

### DIFF
--- a/content/wasm-functions/quickstart.md
+++ b/content/wasm-functions/quickstart.md
@@ -37,7 +37,7 @@ Download the `spin` binary using the `install.sh` script hosted on this site:
 
 <!-- @selectiveCpy -->
 
-<pre class="bash spin-install" id="spin-install-quick"><code>$ curl -fsSL https://wasm-functions.fermyon.app/downloads/install.sh | bash
+<pre class="bash spin-install" id="spin-install-quick"><code>$ curl -fsSL https://developer.fermyon.com/downloads/fwf_install.sh | bash
 </code></pre>
 
 Then move the `spin` binary somewhere in your path, so you can run it from anywhere. For example:
--- a/downloads/fwf_install.sh
+++ b/downloads/fwf_install.sh
@@ -98,7 +98,7 @@ case $OSTYPE in
     ;;
 *)
     fancy_print 1 "The OSTYPE: ${OSTYPE} is not supported by this script."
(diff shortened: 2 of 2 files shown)


content/wasm-functions/quickstart.md
```diff
@@ -37,7 +37,7 @@ Download the `spin` binary using the `install.sh` script hosted on this site:
 
 <!-- @selectiveCpy -->
 
-<pre class="bash spin-install" id="spin-install-quick"><code>$ curl -fsSL https://wasm-functions.fermyon.app/downloads/install.sh | bash
```
calebschoepp, line 40: Can you update the few places in `prod-monitoring.yaml` that rely on this old URL? https://github.com/fermyon/neutrino/blob/0976f623b3829d5ca353f5114bf36d119f8b6af2/.github/workflows/prod-monitoring.yaml#L62. If we don't update this our monitors are going to break.

calebschoepp (approved):

downloads/fwf_install.sh
```diff
@@ -0,0 +1,158 @@
...
+fancy_print 0 "Step 5: Installing default plugins"
+./spin plugins update
+./spin plugins install aka --yes
+
+if [[ $VERSION = "canary" ]]; then
+    fancy_print 3 "Warning: You are using canary Spin, but templates use latest stable SDKs."
+    fancy_print 3 "Be sure to update templates to point to 'canary' SDKs."
+fi
+# Direct to quicks-start doc
+fancy_print 0 "You're good to go. Check here for the next steps: https://wasm-functions.fermyon.app/fermyon-wasm-functions/quickstart"
```
itowlson, line 157: Is this still the right URL?

downloads/fwf_install.sh
```diff
@@ -0,0 +1,158 @@
...
+case $OSTYPE in
+"linux-gnu"*)
+    OS="linux"
+    ;;
+"darwin"*)
+    OS="macos"
+    ;;
+*)
+    fancy_print 1 "The OSTYPE: ${OSTYPE} is not supported by this script."
+    fancy_print 2 "Please refer to this article to install Spin: https://wasm-functions.fermyon.app/fermyon-wasm-functions/quickstart"
```
itowlson, line 101: URL?

PR #1576, opened 2025-08-26
update docs to be compatible with the JS sdk from spinframework

karthik2804:
Updates the docs in `wasm-functions` to be compatible with the modular JS SDK in the Spinframework org.


macolso (approved):
LGTM - thank you for updating these

PR #1615, opened 2026-05-06
ci(deploy-preview.yml): fetch cloud plugin

vdice:


vdice:
~~Requesting an increased max app count for the preview deploy user, looks like it was reverted to the default of 5...~~

karthik2804 (approved):

PR #1619, opened 2026-06-11
Viva la 4.0

fibonacci1729:
Putting this up now but won't merge until we're ready.


itowlson:
Is this the script that the AF docs link to?  If so, should we update it only in sync with the AF docs?

fibonacci1729:
Yeah @itowlson That's the plan 👍

itowlson (approved):

